PR #7, opened 2025-07-08
feat: add workspace

TomasArrachea:
This PR adds the initial workspace structure with all the needed files.

All the faucet related code was moved to `/bin/faucet`. Will be refactored in a separate PR into crates (https://github.com/0xMiden/miden-faucet/issues/6).


README.md
bobbinth, line 1: Could we add the `License` section as we have in other repos?

CONTRIBUTING.md
bobbinth, line 1: Could we use the same file from `miden-base`'s `next` branch. I think it has some minor updates compared to this one.

.dockerignore
bobbinth, line 1: As mentioned in the other PR, should we maybe drop support for Docker for now?

Makefile
```diff
@@ -0,0 +1,105 @@
...
+test:  ## Runs all tests
+	cargo nextest run --all-features --workspace
+
+# --- checking ------------------------------------------------------------------------------------
+
+.PHONY: check
+check: ## Check all targets and features for errors without code generation
+	${BUILD_PROTO} cargo check --all-features --all-targets --locked --workspace
+
+# --- installing ----------------------------------------------------------------------------------
```
bobbinth, line 81: Could we also add tooling install commands - similar to the ones in [miden-base](https://github.com/0xMiden/miden-base/blob/next/Makefile#L143).

Makefile
```diff
@@ -0,0 +1,105 @@
...
+toml-check: ## Runs Format for all TOML files but only in check mode
+	taplo fmt --check --verbose
+
+.PHONY: typos-check
+typos-check: ## Runs spellchecker
+	typos
+
+.PHONY: workspace-check
+workspace-check: ## Runs a check that all packages have `lints.workspace = true`
+	cargo workspace-lints
```
bobbinth, line 53: Similar to the comment in the other PR - I wonder if we can get rid of this.

bobbinth (commented):
Looks good! Thank you! Not a review yet, just left some comments for repo setup related files inline.

CONTRIBUTING.md
```diff
@@ -0,0 +1,134 @@
+# Contributing to Miden Base
+
+#### First off, thanks for taking the time to contribute!
+
+We want to make contributing to this project as easy and transparent as possible, whether it's:
+
+- Reporting a [bug](https://github.com/0xMiden/miden-base/issues/new?assignees=&labels=bug&projects=&template=1-bugreport.yml)
+- Taking part in [discussions](https://github.com/0xMiden/miden-base/discussions)
+- Submitting a [fix](https://github.com/0xMiden/miden-base/pulls)
+- Proposing new [features](https://github.com/0xMiden/miden-base/issues/new?assignees=&labels=enhancement&projects=&template=2-feature-request.yml)
```
SantiagoPittella, line 10: The title and links here are for the miden-base repo. I guess this should be changed to miden-faucet.

Cargo.toml
```diff
@@ -0,0 +1,66 @@
...
+# miden-base aka protocol dependencies. These should be updated in sync.
+miden-lib     = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+miden-objects = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
+miden-testing = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+miden-tx      = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
+
+# Other miden dependencies. These should align with those expected by miden-base.
+miden-air = { version = "0.15" }
+rand      = { version = "0.9" }
+thiserror = { default-features = false, version = "2.0" }
```
SantiagoPittella, line 38: Should we create an issue to re-export them from there?

README.md
```diff
@@ -1,2 +1,6 @@
-# miden-faucet
+# Miden faucet
 Token faucet application for Miden testnet
+
+## License
+
+This project is [MIT licensed](./LICENSE).
```
SantiagoPittella, line 6: The license is mentioned here, but the file doesn't exists. I think that we should copy the MIT license from the other repo cc @bobbinth

bin/faucet/README.md
```diff
@@ -0,0 +1,53 @@
+# Miden node
```
SantiagoPittella, line 1:
```suggestion
# Miden faucet
```

README.md
TomasArrachea: The license is not added in this PR but it's already on the repo [here](https://github.com/0xMiden/miden-faucet/blob/main/LICENSE)

SantiagoPittella (approved):

Cargo.toml
TomasArrachea: I feel like the approach in general with this is to avoid re-exporting deps. For example, we have this same issue on miden-client depending on node and base, instead of the node reexporting base. So I'd say its better to keep that direction for now

bin/faucet/README.md
```diff
@@ -0,0 +1,53 @@
...
+  - **API Keys**: the faucet is initialized with a set of API Keys that can be distributed to developers. The difficulty of the challenges requested using the API Key will increase only with the load of that key, it won't be influenced by the overall load of the faucet.
+
+2. **Requests batching**:
+  - Maximum batch size: 100 requests
+  - Requests are processed in batches to optimize performance
+  - Failed requests within a batch are handled individually
+
+3. **Account rollbacks**:
+  - Faucet maintains the last 1000 account states for potential rollbacks
+  - Is used in case a desync is detected
```
bobbinth, line 49: nit: this list doesn't get rendered properly. Instead of unordered bullet points we should use the ordered ones. That is, instead of:
```
1. abc
  - xyz 
```
We should use:
```
1. abc
  a. xyz.
```

bin/faucet/build.rs
bobbinth, line 1: I wonder if we actually need this for the faucet. I'd keep it for now but let's create an issue to consider removing it.

cc @Mirko-von-Leipzig.

bin/faucet/Cargo.toml
```diff
@@ -0,0 +1,64 @@
+[package]
+authors.workspace      = true
+description            = "Miden node token faucet"
```
bobbinth, line 3: I would change the description to: "Token faucet application for Miden testnet"

Cargo.toml
```diff
@@ -0,0 +1,66 @@
...
+miden-node-proto           = { branch = "next", git = "https://github.com/0xMiden/miden-node" }
+miden-node-rpc             = { branch = "next", git = "https://github.com/0xMiden/miden-node" }
+miden-node-utils           = { branch = "next", git = "https://github.com/0xMiden/miden-node" }
+miden-remote-prover-client = { branch = "next", git = "https://github.com/0xMiden/miden-node" }
+
+# miden-base aka protocol dependencies. These should be updated in sync.
+miden-lib     = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+miden-objects = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
+miden-testing = { branch = "next", git = "https://github.com/0xMiden/miden-base" }
+miden-tx      = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
```
bobbinth, line 33: Let's update these to v0.10 dependencies from crates.io for now.

Not for this PR: ideally, we'd replace these with a single `miden-client` dependency.

Cargo.toml
```diff
@@ -0,0 +1,66 @@
...
+
+# Other miden dependencies. These should align with those expected by miden-base.
+miden-air = { version = "0.15" }
```
bobbinth, line 36: What do we use `miden-air` dependency for? Not for this PR, but we should get rid of it (let's create an issue for this).

bin/faucet/Cargo.toml
```diff
@@ -0,0 +1,64 @@
...
+serde_json                 = { version = "1.0" }
+sha3                       = { version = "0.10" }
+thiserror                  = { workspace = true }
+tokio                      = { features = ["fs"], workspace = true }
+tokio-stream               = { features = ["net"], workspace = true }
+tonic                      = { features = ["tls-native-roots"], workspace = true }
+tower                      = { workspace = true }
+tower-http                 = { features = ["cors", "set-header", "trace"], workspace = true }
+tracing                    = { workspace = true }
+url                        = { workspace = true }
```
bobbinth, line 45: Let's categorize the dependencies into workspace/miden/external dependencies.

Makefile
```diff
@@ -0,0 +1,97 @@
...
+
+# --- docs ----------------------------------------------------------------------------------------
+
+.PHONY: doc
+doc: ## Generates & checks documentation
+	$(WARNINGS) cargo doc --all-features --keep-going --release --locked
+
+.PHONY: book
+book: ## Builds the book & serves documentation site
+	mdbook serve --open docs
```
bobbinth, line 62: We don't currently have the mdBook set up for the faucet, right? If so, I'd remove this from here and create an issue to write mdBook docs in the future.

Cargo.toml
TomasArrachea: Seems it was not needed so I removed it

bin/faucet/README.md
TomasArrachea: markdown only supports numbered ordered lists, not with letters, so instead changed it to be all unordered

bin/faucet/Cargo.toml
```diff
@@ -0,0 +1,69 @@
...
+humantime    = { workspace = true }
+rand         = { features = ["thread_rng"], workspace = true }
+thiserror    = { workspace = true }
+tokio        = { features = ["fs"], workspace = true }
+tokio-stream = { features = ["net"], workspace = true }
+tonic        = { features = ["tls-native-roots"], workspace = true }
+tower        = { workspace = true }
+tower-http   = { features = ["cors", "set-header", "trace"], workspace = true }
+tracing      = { workspace = true }
+url          = { workspace = true }
```
bobbinth, line 30: I think these are also external dependencies.

PR #8, opened 2025-07-08
feat: add CI workflows

TomasArrachea:
This PR ports the CI checks from miden-node into this repo.

PR #10, opened 2025-07-10
feat: separate faucet into crates

TomasArrachea:


bobbinth:
Once #7 and #8 are merged, all subsequent PRs (starting with this one) should probably go into the `next` branch.

crates/backend/README.md
```diff
@@ -0,0 +1,11 @@
+# Miden Faucet Backend
+
+This crate provides the backend services for the Miden Faucet. This includes handling token minting, proof-of-work challenges, and API endpoints.
+
+## API Endpoints
+
+- `GET /api/pow` - Request a proof-of-work challenge
+- `GET /api/get_tokens` - Request tokens
```
tomyrd, line 8: We should specify the endpoint parameters

crates/backend/src/server/get_tokens.rs
```diff
@@ -170,7 +166,12 @@ impl RawMintRequest {
             .ok_or(MintRequestError::AssetAmount(self.asset_amount))?;
 
         // Check the API key, if provided
-        let api_key = self.api_key.as_deref().map(ApiKey::decode).transpose()?;
+        let api_key = self
+            .api_key
+            .as_deref()
+            .map(ApiKey::decode)
+            .transpose()
+            .map_err(|_| MintRequestError::InvalidApiKey(self.api_key.unwrap_or_default()))?;
```
tomyrd, line 174: If we end up ignoring the anyhow error to convert it back to `MintRequestError`, shouldn't we leave it as before? Where `ApiKey::decode` returned a `MintRequestError`

tomyrd (approved):
LGTM, left a couple of small comments

crates/backend/src/server/get_tokens.rs
TomasArrachea: The API key is decoded in both the `pow` and `get_tokens` endpoints so I think it should return a generic error and be mapped on each endpoint. However, there was actually a naming error in the PoW endpoint so I fixed that in [734e149](https://github.com/0xMiden/miden-faucet/pull/10/commits/734e149d0b4903968ff2a1364039feeb20b22201)

TomasArrachea:
> I don't think the website needs to be it's own crate

I agree, we can actually make the website a binary crate that holds the static files. Maybe even try using [`axum_http::ServeDir`](https://docs.rs/tower-http/latest/tower_http/services/struct.ServeDir.html) to serve them. This also makes sense to be able run/deploy the frontend separately from the server. The backend address could be set on a `config.json` file.

> The faucet part could be a library crate.

Yes, we can definitely separate it to a `miden-faucet` crate, that would be quite small and depend on `miden-client`.

> Another option could be to make miden-faucet into a binary

I think this may be the more intuitive design. The `miden-faucet` binary would expose the CLI to init/import the faucet, but not have a `start` command, since it would be run by the server. The server would start the faucet, connect with it through channels, and expose the backend endpoints.

In summary, the structure would look like:
- `miden-faucet` binary crate
- `server` binary
- `website` binary

As a downside, this would require that to run the faucet we need to interact with all three binaries, and the frontend and backend would be run separately. Maybe it could be simplified with some makefile targets.

> we should probably refactor this struct to enable handling of both read and write requests

Read requests would be to query the db for the notes? The problem with this is that it would limit the concurrency for reading, since the Store is Sync+Send but the client isn't.

bobbinth:
> In summary, the structure would look like:
> 
> * `miden-faucet` binary crate
> * `server` binary
> * `website` binary

Overall, makes sense - though, I'm not 100% sure about the website binary - maybe there are easier ways to serve static content?

A much more radical approach, which we probably shouldn't try but wanted to bring up nevertheless, is to use something like [Deno](https://deno.com/) for both server and website. I've never used it before but I have been following the project for a while, and my understanding is that it integrates with Rust pretty well. The main benefit would be that the majority of logic could be written in TypeScript and so would have a much broader potential developer base. Again, probably not something to try now, but let's keep this in mind (and maybe do some experiments along the way). cc @mmagician and @Dominik1999.

> Read requests would be to query the db for the notes? The problem with this is that it would limit the concurrency for reading, since the Store is Sync+Send but the client isn't.

Yes, not something we can do right now, but I think we are moving toward making the client Send+Sync, and once that becomes available, we could make use of it.

mmagician (approved):
I think the split looks good, we can always make follow-up refactors if needed.

Some deployment things to keep in mind:
1. The `Makefile` needs to be updated to reflect the split (I think this is just a convenience command but still), and
2. The binary that needs to run is no longer called `miden-faucet` but `server`, correct? Then I think the {test/dev}net deployment jobs need to be updated. @sergerad

bobbinth:
> I've come to think that with this split its best to just keep the binary name as `bin/faucet` and name the crate as `miden-faucet-client`. This way makes more sense for the binary to hold the CLI with all the faucet commands, including the next refactor #34.

Could you explain a bit more how you are thinking about this? i.e., what would `miden-faucet-client` binary do? would there be another binary to run the faucet?

TomasArrachea:
At the moment, this PR just extracts the faucet logic into a new library crate: `miden-faucet-client`. This crate is basically the wrapper of miden-client that runs the faucet loop for handling the minting requests.

The `miden-faucet` binary still holds the CLI and the server implementation. I feel that is better to have a single binary with the CLI that interacts with both the server and the faucet. Otherwise, starting the faucet would require interacting with multiple binaries and seems confusing.

I'd say that we can discuss separating the server or frontend from the `miden-faucet` binary as a separate issue. I'm still not sure about how would it be beneficial.

bobbinth (commented):
Thank you! Looks good. I left some comment inline. The main ones about better separation between the library and the binary (i.e., the library ideally should not assume anything about how it is used down stream) and also about the name for the library crate.

sergerad:
> The binary that needs to run is no longer called miden-faucet but server, correct? Then I think the {test/dev}net deployment jobs need to be updated.

This would break our infra automation I would recommend not making those kinds of changes at all.

crates/client/src/types.rs
```diff
@@ -120,3 +119,18 @@ impl std::fmt::Display for NoteType {
...
+#[derive(Debug, Clone, Serialize)]
+pub struct ExplorerUrl(&'static str);
+
+impl ExplorerUrl {
+    /// Returns the explorer URL for the given network ID.
+    /// Currently only testnet explorer is available.
+    pub fn from_network_id(network_id: NetworkId) -> Option<Self> {
+        match network_id {
+            NetworkId::Testnet => Some(Self("https://testnet.midenscan.com")),
+            _ => None,
```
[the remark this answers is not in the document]
TomasArrachea: https://github.com/0xMiden/miden-faucet/issues/48

bin/faucet/src/static/index.html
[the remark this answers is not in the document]
TomasArrachea: Yes, they can. Moved them there

crates/client/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+authors.workspace      = true
+description            = "Miden faucet's logic"
+edition.workspace      = true
+homepage.workspace     = true
+keywords               = ["faucet", "miden", "node"]
+license.workspace      = true
+name                   = "miden-faucet-client"
```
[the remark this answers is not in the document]
TomasArrachea: I went with renaming the crate to `miden-faucet-lib`, so we avoid renaming the binary and breaking the infra.

crates/client/src/requests.rs
```diff
@@ -0,0 +1,64 @@
...
+    ServerSignaturesDoNotMatch,
+    #[error("server timestamp expired, received: {0}, current time: {1}")]
+    ExpiredServerTimestamp(u64, u64),
+    #[error("challenge already used")]
+    ChallengeAlreadyUsed,
+    #[error("account is rate limited")]
+    RateLimited,
+    #[error("faucet supply exceeded")]
+    AvailableSupplyExceeded,
+}
```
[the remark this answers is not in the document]
TomasArrachea: Right, I refactored this into separate errors

crates/client/src/lib.rs
```diff
@@ -69,7 +72,14 @@ pub struct Faucet {
...
+    /// Loads the faucet.
+    ///
+    /// A client is instantiated with the provided store path, node url and timeout. The account is
+    /// loaded from the provided account file. If the account is already tracked by the current
+    /// store, it is loaded. Otherwise, the account is added from the file state.
+    ///
+    /// If a remote transaction prover url is provided, it is used to prove transactions. Otherwise,
+    /// a local transaction prover is used.
     #[instrument(name = "faucet.load", fields(id), skip_all)]
     pub async fn load(
```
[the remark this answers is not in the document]
TomasArrachea: I'll introduce this changes in a follow-up PR to avoid delaying this one, it might trigger some discussion.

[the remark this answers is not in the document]
TomasArrachea: We can join this work with https://github.com/0xMiden/miden-faucet/issues/34, since that issue will involve a refactor on the `Faucet::load` function.

bin/faucet/src/api/frontend.rs
```diff
@@ -0,0 +1,31 @@
+use axum::response::{Html, IntoResponse, Response};
+use axum_extra::response::{Css, JavaScript};
+use http::header::{self};
+
+pub async fn get_index_html() -> Html<&'static str> {
+    Html(include_str!("../../frontend/index.html"))
+}
```
bobbinth, line 7: Could we make the path here relevant to something like `CARGO_MANIFEST_DIR`? For example:
```rust
Html(include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/frontend/index.html")))
```

Also, would be good to confirm that this will work fine during publishing (I think it should work, but wanted to double-check).

bin/faucet/src/api/frontend.rs
bobbinth, line 1: Question (not for this PR): is there a way to just serve everything from the `/frontend` directory (i.e., without having to create a separate function for each file)?

crates/client/README.md
```diff
@@ -0,0 +1,9 @@
+# Miden Faucet Client
```
bobbinth, line 1: I would just call this "Miden faucet"

crates/client/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+This crate implements a faucet to be used by the website. The faucet uses `miden-client` to create, execute, prove transactions, and then submits them to the node.
+
+The faucet is supposed to be run in a single thread, and receives minting requests via channels.
+
+## License
+
+This project is licensed under the MIT License - see the LICENSE file for details.
```
bobbinth, line 9: Could we use the same phrasing here as we do in other crates?

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["bin/faucet"]
+members = ["bin/faucet", "crates/client"]
```
bobbinth, line 2: Not sure about using "client" here still. I think we could also just call it "faucet". So, this line could be:
```toml
members = ["bin/faucet", "crates/faucet"]
```
Or maybe
```toml
members = ["bin/faucet", "crates/faucet-lib"]
```

bin/faucet/src/pow/mod.rs
bobbinth, line 1: Not for this PR, but would be good to investigate if we can break out all PoW-related functionality into a separate `utils` create (e.g., `miden-faucet-utils`).

bobbinth (approved):
Looks good! Thank you! I left some more comments inline - but they are pretty small. After the comments are addressed, we can merge.

PR #11, opened 2025-07-11
feat: integrate miden client

TomasArrachea:
Part of https://github.com/0xMiden/miden-faucet/issues/9

This PR integrates miden-client as a dependency and uses it to execute, prove, and submit transactions. Since the client handles the proving and submission in a single funcion, the `Proven` update was removed. 

There is some more refactors to be done in a following PR:
- Sync the client with the node to track the commited notes
- Send the commited notes update to the frontend. We probably need to add a separate endpoint for this.


bin/faucet/Cargo.toml
```diff
@@ -53,9 +52,10 @@ miden-node-utils = { features = ["vergen"], workspace = true }
 
 [dev-dependencies]
 fantoccini                = { version = "0.22" }
-miden-node-block-producer = { features = ["testing"], workspace = true }
+miden-node-block-producer = { features = ["testing"], version = "0.10.0" }
+miden-node-proto          = { version = "0.10.0" }
 miden-objects             = { features = ["testing"], workspace = true }
-miden-testing             = { features = ["async"], workspace = true }
```
tomyrd, line 58: Same as above

bin/faucet/src/faucet/mod.rs
```diff
@@ -281,177 +180,59 @@ impl Faucet {
...
-                    && err.message().contains("incorrect initial state commitment") =>
-            {
-                self.handle_desync(err.message()).with_context(|| {
-                    format!("failed to recover from desync error: {}", err.message())
-                })
-            },
-            // TODO: Look into which other errors should be recoverable.
-            //       e.g. Connection error being lost to RPC client is probably not fatal.
-            others => Err(others).context("failed to handle error"),
+            let block_height = self.client.get_sync_height().await?;
```
tomyrd, line 185: Just to be clear, this call will return the client's block height, not the node. Given that we are not updating the client with `state_sync`, this will always return 0.

tomyrd (approved):
Looks good, left a couple of minor comments.

CHANGELOG.md
```diff
@@ -1 +1,7 @@
 # Changelog
+
+## 0.1.0 (TBD)
+
+### Changes
+
+- Integrated miden-client (#11).
```
SantiagoPittella, line 7: The package already exists, we should continue it. Maybe with something like
```suggestion

## 0.11.0 (TBD)

### Changes

- Integrated miden-client (#11).

<a message with something like "For previous version go to miden-node changelog">
```

bin/faucet/src/stub_rpc_api.rs
TomasArrachea, line 1: This stubbed API may be replaced with the `MockRpcApi` from miden-client. I'm leaving this for a separate PR since it needs some refactors in the startup to somehow inject the mocked client.

bin/faucet/src/server/challenge.rs
```diff
@@ -162,6 +160,25 @@ impl Challenge {
...
+    /// Simple hex decoding function.
+    fn hex_decode(s: &str) -> Result<Vec<u8>, ()> {
+        if s.len() % 2 != 0 {
+            return Err(());
+        }
+        (0..s.len())
+            .step_by(2)
+            .map(|i| u8::from_str_radix(&s[i..i + 2], 16).map_err(|_| ()))
+            .collect()
+    }
```
tomyrd, line 181: Should we use an [external crate for hex encoding](https://docs.rs/hex/latest/hex/)? If not, we should add some tests to ensure our implementation works correctly

tomyrd (approved):
LGTM

CHANGELOG.md
```diff
@@ -1 +1,11 @@
...
+
+## 0.11.0 (TBD)
+
+### Changes
+
+- Integrated miden-client (#11).
+
+## 0.10.0 (2025-07-10)
+
+For previous changes see https://github.com/0xMiden/miden-node/blob/main/CHANGELOG.md
```
igamigo, line 11: I'm not sure we want to reference the previous log (cc @bobbinth for an opinion). But if we do, probably no need to reference `main` because it will eventually show all the other newer versions which did not host the faucet. So maybe you can reference a specific commit instead.

bin/faucet/src/faucet/mod.rs
```diff
@@ -133,133 +65,97 @@ pub struct MintRequest {
...
-
-        tracing::Span::current().record("id", id.account_id.to_string());
+        let account = account_file.account;
+        tracing::Span::current().record("id", account.id().to_string());
+
+        let keystore = FilesystemKeyStore::<StdRng>::new(PathBuf::from("keystore"))
+            .context("failed to create keystore")?;
+        keystore.add_key(
+            account_file.auth_secret_keys.first().context("account file has no auth keys")?,
+        )?;
```
igamigo, line 94: Maybe wouldn't hurt to add all secret keys to the keystore.

bin/faucet/src/faucet/mod.rs
```diff
@@ -133,133 +65,97 @@ pub struct MintRequest {
...
+        let endpoint = Endpoint::try_from(node_url.as_str())
+            .map_err(|e| anyhow::anyhow!("failed to parse node url: {e}"))?;
```
igamigo, line 96: nit:
```suggestion
            .with_context(|| format!("invalid node url: {}", node_url))?;
```

bin/faucet/src/faucet/mod.rs
```diff
@@ -281,177 +177,51 @@ impl Faucet {
...
+        tracing::info!("Request stream closed, shutting down minter");
 
         Ok(())
     }
 
-    /// Fully handles a batch of requests _without_ changing local state.
-    ///
-    /// Caller should update the local state based on the returned result.
+    /// Fully handles a batch of requests to create and submit a transaction, and updates the local
+    /// db.
```
igamigo, line 192: I would explain here what this involves (for example, getting multiple requests and submitting notes for them, updating clients in the meantime, updating the local Miden client, etc.)

bin/faucet/src/faucet/mod.rs
```diff
@@ -463,75 +233,51 @@ impl Faucet {
...
-    ) -> Result<Self, MintError> {
-        // If building a note fails, we discard the whole batch. Should never happen, since account
-        // ids are validated on the request level.
-        let mut notes = Vec::new();
-        for request in requests {
-            let amount = request.asset_amount.inner() * 10u64.pow(decimals.into());
-            // SAFETY: source is definitely a faucet account, and the amount is valid.
-            let asset = FungibleAsset::new(source.account_id, amount).unwrap();
-            let note = create_p2id_note(
+/// Builds a collection of `P2Id` notes from a set of mint requests.
```
igamigo, line 236: nit: 
```suggestion
/// Builds a collection of `P2ID` notes from a set of mint requests.
```

bin/faucet/src/main.rs
```diff
@@ -285,17 +292,18 @@ async fn run_faucet_command(cli: Cli) -> anyhow::Result<()> {
...
-                Felt::try_from(max_supply)
-                    .expect("max supply value is greater than or equal to the field modulus"),
-                AccountStorageMode::Public,
-                AuthScheme::RpoFalcon512 { pub_key: secret.public_key() },
-            )
-            .context("failed to create basic fungible faucet account")?;
+            let symbol = TokenSymbol::try_from(token_symbol.as_str())
+                .context("failed to parse token symbol")?;
+            let max_supply = Felt::try_from(max_supply)
+                .expect("max supply value is greater than or equal to the field modulus");
```
igamigo, line 298:
```suggestion
                .context("max supply value is greater than or equal to the field modulus")?;
```

bin/faucet/src/main.rs
```diff
@@ -523,6 +532,7 @@ mod test {
                         faucet_account_path: faucet_account_path.clone(),
                         remote_tx_prover_url: None,
                         open_telemetry: false,
+                        store_path: PathBuf::from("store.sqlite3"),
```
igamigo, line 535: Let's name this `faucet_client_store.sqlite3` or something more representative

bin/faucet/src/server/challenge.rs
igamigo: Is the problem that we don't expose the hex utils (`miden_tx::utils::{ToHex, hex_to_bytes};`) from the client? If so, I believe we could re-export them unless there's a good reason not to. Could you create an issue/PR?

igamigo (approved):
LGTM! Left some comments, mostly nits

bin/faucet/src/main.rs
TomasArrachea: This error type is a String so it doesn't implement StdError and we can't use anyhow context on it

bin/faucet/src/faucet/mod.rs
TomasArrachea: Same comment about this error being a String

bin/faucet/src/main.rs
```diff
@@ -12,17 +11,19 @@ use std::{num::NonZeroUsize, path::PathBuf, time::Duration};
...
+    account::{
+        AccountBuilder, AccountFile, AccountStorageMode, AccountType,
+        component::{BasicFungibleFaucet, RpoFalcon512},
+    },
     asset::TokenSymbol,
-    crypto::dsa::rpo_falcon512::SecretKey,
+    auth::AuthSecretKey,
+    crypto::SecretKey,
 };
+use miden_node_utils::{crypto::get_rpo_random_coin, logging::OpenTelemetry, version::LongVersion};
```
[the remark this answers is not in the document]
TomasArrachea: I think maybe we could just remove the `LongVersion`. Not sure why it was added in the node but maybe it's not required for the faucet? cc @Mirko-von-Leipzig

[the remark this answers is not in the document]
TomasArrachea: Its basically this file: https://github.com/0xMiden/miden-node/blob/next/crates/utils/src/logging.rs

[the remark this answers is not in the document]
bobbinth: Let's move this file here (probably in a separate PR) - we can also probably simplify it a bit.

PR #19, opened 2025-07-17
feat: sync miden client and expose commited notes

TomasArrachea:
Closes #11 

This PR adds a new `get_notes` endpoint that accesses the faucet store to retrieve notes. This replaces the previous `Minted` sse event that sent the note to the frontend once it was built. 

The frontend was updated to add a new `Download Note` button that hits this endpoint to get the created note. Nevertheless, it will be refactored in the new frontend redesign. Currently looks like:
image

Note: this PR is implemented on top of #11 that integrates the client


bin/faucet/src/faucet/updates.rs
```diff
@@ -1,112 +1,62 @@
...
-impl MintUpdate<'_> {
+impl MintUpdate {
     /// Converts the mint update into an sse event.
     /// Event types:
     /// - `MintUpdate::Built`: event type "update"
     /// - `MintUpdate::Executed`: event type "update"
     /// - `MintUpdate::Submitted`: event type "update"
-    /// - `MintUpdate::Minted`: event type "note". Contains the note encoded in base64 if it is
+    /// - `MintUpdate::Minted`: event type "minted". Contains the note encoded in base64 if it is
     ///   private.
```
tomyrd, line 49: This comment should be updated as we no longer encode the full note in the update, only the ID.

bin/faucet/src/server/get_notes.rs
```diff
@@ -0,0 +1,96 @@
...
+    let request = request.validate()?;
+    let note = server
+        .store
+        .get_output_notes(NoteFilter::Unique(request.note_id))
+        .await
+        .map_err(|e| {
+            tracing::error!("failed to read note from store: {}", e);
+            NoteRequestError::NoteNotFound
+        })?
+        .pop()
```
tomyrd, line 40: If the endpoint always returns one note it should be called `get_note`, if not we should remove this `pop` and receive an array of requested `note_id`.

bin/faucet/src/server/get_notes.rs
```diff
@@ -0,0 +1,96 @@
...
+// ENDPOINT
+// ================================================================================================
+
+#[instrument(
+    parent = None, target = COMPONENT, name = "faucet.server.get_notes", skip_all,
+    fields(
+        note_id = %request.note_id,
+    )
+)]
+pub async fn get_notes(
```
tomyrd, line 27: We should add the new request to the README

bin/faucet/src/stub_rpc_api.rs
bobbinth, line 1: nit: I would maybe move this file under `testing` directory. For example: `src/testing/stub_rpc.rs` (and we'd also need `src/testing/mod.rs` to export it).

bin/faucet/src/faucet/updates.rs
```diff
@@ -1,112 +1,62 @@
...
+impl MintUpdate {
     /// Converts the mint update into an sse event.
     /// Event types:
     /// - `MintUpdate::Built`: event type "update"
     /// - `MintUpdate::Executed`: event type "update"
     /// - `MintUpdate::Submitted`: event type "update"
-    /// - `MintUpdate::Minted`: event type "note". Contains the note encoded in base64 if it is
-    ///   private.
+    /// - `MintUpdate::Minted`: event type "minted". Contains the note id, transaction id and
+    ///   explorer url.
```
bobbinth, line 49: Is "explorer ur" here correct? Seems like we are sending just the `NetworkId` - right?

bobbinth (approved):
Looks good! Thank you! Not a very deep review from me - but I left a couple of small comments inline.

bin/faucet/src/faucet/updates.rs
TomasArrachea: Yes, the `NetworkId` is actually used to generate the explorer url

Issue #13, opened 2025-07-12
Update token request flows


bobbinth:
We should update token request flows to look similar to the sketches below.

<img width="14505" height="4332" alt="Image" src="https://github.com/user-attachments/assets/94f9d620-8e4d-40ad-b899-4976cb57b612" />

<img width="14502" height="4332" alt="Image" src="https://github.com/user-attachments/assets/0464808f-7b25-439a-887a-543c4473982f" />

The request for tokens should go through the following stages, which should be communicated via the progress indicator. The stages are:

1. Performing the proof-of-work. The text for this could be "Preparing the request".
2. Once the request is sent to the server, the text should change to something like: "Minting tokens" (this is what is in the images right now).
3. Once the proven transaction has been sent to the network but before the transaction was committed, the text should be something like: "Waiting for confirmation".
4. Once the transaction has been committed we should show the 3rd image in the designs - but the "View in explorer" button should be disabled.
5. Once the note becomes available in the explorer, the "View in explorer" button should become enabled.

To implement this, we probably will need to re-work quite a few of the faucet's internals which may require various sub-issues.

bobbinth:
Tagging @dotsolo for any design-related questions.

TomasArrachea:
Using the "read-only" client to fetch commited notes from the db solves (temporarily) how to send the note to the frontend when requested with the download button. Another problem arises from being able to only show the button as available only after the note has been committed and the client received the committed update in the sync. I can think of some possible paths forward (there may be more solutions): 
1. The "download note" button is always available. The frontend receives the note without inclusion proof that is read from the miden-client store by the read-only client.
2. The "download note" button is always available, but the backend returns an error if the note has not yet been commited.
3. The button is unavailable, and we use the web-client to poll the node for the note id. Once is commited, we enable the button.
4. Some other solution, like creating a new server-side event that sends an update to the frontend when a note has been commited.

For now I would go for option 1, and create a separate issue to discuss how to enable/disable the download button.

bobbinth:
Yes, the simpler solution is fine for now. My understanding is that the user will be able to download the note (w/o the inclusion proof) right after the transaction is sent to the node, right?

In the future, we can improve on this.

TomasArrachea:
> Yes, the simpler solution is fine for now. My understanding is that the user will be able to download the note (w/o the inclusion proof) right after the transaction is sent to the node, right?

Correct.

PR #20, opened 2025-07-18
feat: main page redesign

### LINKED ISSUE #12: Main page re-design
We should update the main page of the faucet to look similar to the one below (let me know if we need the design in some other format).

-----

Image

-----

The only new element here, as compared to the current faucet is the total supply and number of claimed tokens on the bottom right. This could be implemented over time as follows:

1. We can skip this element for this issue.
2. We'll have a separate issue to show this element statically (i.e., refreshed only when the page is refreshed).
3. And then we'll have another issue (much lower priority) to have this element updated dynamically.

------

Another cool thing, but probably for a different issue, would be to detect if Miden wallet is installed by the browser and show something like "Select" button in the right part of the "Recipient Address" box. Pressing this button would open up the wallet and let the user select one of their accounts.

cc

TomasArrachea:
closes #12 

This PR redesigns the homepage of the faucet frontend while keeping it working with the same backend interface.

Note: 
- The design showed in the issue has been updated, this is the latest version: https://www.figma.com/design/CKvKBn0VtSFSvB6s4z5X62/Miden-Faucet?node-id=0-1&t=HoXL8xBO3bQ0ArLL-1
- in a following PR the status updates will be refactored so for now they are being showed in a temporary message.
- also need to add the tokens claimed and available to the metadata request to be able to show it in the frontend, leaving hardcoded values for now.

Preview:
image


bin/faucet/src/server/resources/index.js
```diff
@@ -229,7 +241,6 @@ document.addEventListener('DOMContentLoaded', function () {
                 // Take the first 8 bytes of the hash and parse them as u64 in big-endian
                 let digest = BigInt("0x" + hash.hex().slice(0, 16));
 
-                // Check if the hash is less than the target
```
igamigo, line 232: I would leave these comments, and maybe even expand them to make it more comprehensive

igamigo (approved):
Looks great! Tagging @bobbinth to see if there are any design-related comments

bin/faucet/src/server/resources/index.js
```diff
@@ -1,206 +1,220 @@
...
+                for (const amount of data.asset_amount_options) {
+                    const option = document.createElement('option');
+                    option.value = amount;
+                    option.textContent = amount;
+                    this.tokenSelect.appendChild(option);
+                }
+                // TODO: add metadata values
+                this.tokensClaimed.textContent = '0';
+                this.tokensSupply.textContent = '100,000,000,000';
+                this.progressFill.style.width = '0%';
```
tomyrd, line 86: Should this TODO be handled in this PR or in a followup?

bin/faucet/src/server/resources/index.js
```diff
@@ -1,206 +1,220 @@
...
+            });
+    }
 
-        // Get the PoW challenge from the new /pow endpoint
+    async getPowChallenge(recipient) {
         let powResponse;
         try {
             powResponse = await fetch(window.location.href + 'pow?' + new URLSearchParams({
-                account_id: accountAddress
+                account_id: recipient
```
tomyrd, line 98: This may be a bit confusing because in miden recipient refers to a separate concept related to notes (its script, serial number and inputs). Although here the context is a bit different so it might not be a big deal.

We could maybe rename it to `walletAddress` to signify that it's the receiving end of the transaction (i.e. not the faucet).

bin/faucet/src/server/resources/index.js
```diff
@@ -1,206 +1,220 @@
...
 
-        const nonce = await findValidNonce(powData.challenge, powData.target);
+        return await powResponse.json();
+    }
 
-        // Build query parameters for the request using new challenge format
+    async getTokens(challenge, nonce, recipient, amount, isPrivateNote) {
         const params = {
-            account_id: accountAddress,
+            account_id: recipient,
```
tomyrd, line 118: Same as above

bin/faucet/src/server/resources/index.js
```diff
@@ -1,206 +1,220 @@
...
+        this.recipientInput.value = '';
+        this.validateForm();
     }
 
-    // Function to find a valid nonce for proof of work using the new challenge format
-    async function findValidNonce(challenge, target) {
+    truncateAddress(address) {
+        if (address.length <= 10) return address;
+        return `${address.slice(0, 6)}...${address.slice(-4)}`;
+    }
```
tomyrd, line 205: Should we document this function? It isn't clear what its use is and what is being truncated.

tomyrd (commented):
Left some comments, I still need to test it locally before approve

bobbinth (approved):
Looks great! Thank you! Though, this is a pretty shallow review from me since is been a very long time since I did anything HTML/JS/CSS-related.

Two questions:
- This is currently going into `main` - is this intentional? If so, is the idea that we'd deploy this to the testnet sooner rather than later?
- Is there a away to deploy this to some staging server? Maybe we should create a devent server for the faucet? cc @sergerad and @Mirko-von-Leipzig.

bobbinth:
> We do have a separate faucet server for devnet. We would need to:
> 
> * [ ]  Make a faucet release with an attached debian package here.
> * [ ]  Update our infra repo to pull from here.
> * [ ]  Double check that we aren't nuking the faucet account as part of deployment.

Let's create issues for these. Once this is up and running, we could probably remove the faucet code from the `miden-node` repo.

TomasArrachea:
> This is currently going into main

This was just a mistake from my part, I would merge this to next and deploy once the whole frontend redesign is ready (including #13). Nevertheless, being able to use devnet to deploy `next` would be super useful.

tomyrd (changes requested):
Some issues I found while testing locally:

- The success message seems to be a bit messed up, I don't know if it's my window size (I'm using the standard 13' mac screen):

Image

Also, the explorer link should be clickable.

- For private notes I don't have a way to download the note (I think this is being addressed in a separate PR).

- The account address seems to get cleared after minting, is this intended? It is a bit annoying to paste the address again.

- The "Tokens claimed" counter doesn't seem to increment.

TomasArrachea:
@tomyrd Thanks for running it!

> The success message seems to be a bit messed up

Yes, this is just a added as a temporary message since the new design adds a modal for the loading stage (will be added in this follow-up PR #25). 

> For private notes I don't have a way to download the note

This PR didn't change the flow so it's being downloaded automatically once the minting finishes (download button logic is being added in https://github.com/0xMiden/miden-faucet/pull/19).

> The account address seems to get cleared after minting

Yes, this was just a decision I made. I don't expect users to send mint requests in batches so I figured clearing the input makes sense after pressing the button. Can definitely change it though.

> The "Tokens claimed" counter doesn't seem to increment

Yes, I just added placeholder values for now. We have this issue for it: https://github.com/0xMiden/miden-faucet/issues/21.

bin/faucet/src/server/resources/index.js
TomasArrachea: I removed it since it was not actually needed

bin/faucet/src/server/resources/index.js
TomasArrachea: I used this variable to refer to the input of the frontend, since now the form label for the address is shown as: `RECIPIENT ADDRESS`

PR #25, opened 2025-07-22
feat: redesign token request flows

### LINKED ISSUE #13: Update token request flows

TomasArrachea:
closes #13.

This PR redesigns the frontend shown when submitting a mint request.

Also refactors the `get_tokens` endpoint to be a simple GET request, instead of opening a sse channel.

Preview for public note:

https://github.com/user-attachments/assets/f9ae2372-bd02-41af-9635-c4522847381b

Preview for private note:

https://github.com/user-attachments/assets/f9559a4c-2ccb-419a-998d-b71a2bd2cc4b


CHANGELOG.md
```diff
@@ -1 +1,11 @@
 # Changelog
+
+## 0.11.0 (TBD)
+
+### Changes
+
+- Redesigned the home frontend (#20).
```
bobbinth, line 7: Should we add an entry here for the token flows re-redesign?

Also, we've started adding actual links to PR numbers. Let's do it here as well.

bobbinth (approved):
Looks great! Thank you! A pretty shallow review from me, but i left one comment inline.

dotsolo:
> Thanks for the reviews! I still have a few questions about the UX (cc @dotsolo):
> 
> 1. How should the progress bar move? Currently it just moves to 50% on the second step and directly to 100% on the last one (changes on this are probably for a separate PR and low priority).
> 2. After the minting finishes, it shows "CLICK ANYWHERE TO CONTINUE". If a private note was generated, the user has to click on "download" to get the note. Any missclicks would make the user lose access to the note. Perhaps we should replace it with a traditional "x" in the modal's right corner?

Hey @TomasArrachea! Appreciate the notes!

1. For the progress bar animation I was thinking it should simply rotate. This conveys the process is in motion without needing to display a percentage, and it can smoothly transition from that screen to the final one. 

2. Great point and I completely agree. Adding the "x" in the top right corner of the modal is the correct solution. I'll also replace the "CLICK ANYWHERE TO CONTINUE" message. It'll instead be a small confirmation pop-up that occurs after you hit "Download Note" and says "YOUR NOTE HAS BEEN DOWNLOADED. CLICK X TO EXIT"

PR #31, opened 2025-07-29
feat: add supply exceeded check

TomasArrachea:
closes https://github.com/0xMiden/miden-faucet/issues/23.

This PR adds a validation to check that the minting request amount does not exceed the available tokens to claim and the error is correctly showed in the frontend.

### Implementation details
This is done by adding an available supply counter on the faucet. During requests handling, the counter is decremented and the requests that exceed the available supply are filtered. 

A simpler alternative could be to just check for the execution error, but this would need to cancel the whole request batch even if only one request exceeds the limit.


TomasArrachea:
I'm merging this auxiliary PR as I think it's better to review the supply token tracking feature altogether in a single PR.

PR #32, opened 2025-07-29
feat: remove `miden-node-utils` depedency

TomasArrachea:


bin/faucet/src/server/get_tokens.rs
```diff
@@ -137,7 +136,7 @@ impl GetTokenError {
     /// Take care to not expose internal errors here.
     fn user_facing_error(&self) -> String {
         match self {
-            Self::InvalidRequest(invalid_request) => invalid_request.as_report(),
+            Self::InvalidRequest(invalid_request) => invalid_request.to_string(),
```
igamigo, line 139: You'd lose source error information with this direct conversion. Probably ok to keep it this way since it's an invalid request error, though, and this seems like the only place where it's used.

bin/faucet/src/main.rs
```diff
@@ -501,11 +486,11 @@ mod test {
 
         // Wait for faucet to be up
         let endpoint = Url::parse("http://localhost:8080").unwrap();
-        let addr = endpoint.to_socket().unwrap();
+        let addr = endpoint.socket_addrs(|| None).unwrap();
         let start = Instant::now();
         let timeout = Duration::from_secs(10);
         loop {
-            match tokio::net::TcpStream::connect(addr).await {
+            match tokio::net::TcpStream::connect(&*addr).await {
```
igamigo, line 493: nit: I think you can do `&addrs[..]` here instead of `&*addr`. I'd prefer the first one because it's a bit more clear than the latter, that relies a bit on auto dereferencing

igamigo (approved):
LGTM! Left two very minor comments

Mirko-von-Leipzig (commented):
Looks fine to me :) We'll just need to keep an eye on potential changes in the tracing setup in the future. Might be worth creating a standalone crate for just that.

bin/faucet/src/server/get_tokens.rs
TomasArrachea: All the other errors contain no inner error so they would stay the same

bobbinth (approved):
Looks good! Thank you! Though, this is not a very deep review from me.

bin/faucet/src/server/get_tokens.rs
Mirko-von-Leipzig: Ideally the error would indicate why its invalid. In this instance I agree its a bit difficult to understand, but having additional information is useful imo.

```
Failed to receive tokens: account ID parsing failed
  caused by: invalid bech32 account ID length
```

cc @PhilippGackstatter maybe the bech32 error chain could be simplified?

PR #35, opened 2025-07-31
chore: update `miden-client` to `f475d89f` and `mide-base` to `871ac9`

mmagician:
some other dependencies like miden-crypto were also updated

instead of downgrading https://github.com/0xMiden/miden-client/pull/1105, we should now have all versions matching

closes https://github.com/0xMiden/miden-faucet/pull/33

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -400,9 +400,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.42"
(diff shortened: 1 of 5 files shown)


Mirko-von-Leipzig (approved):

mmagician:
Tested locally, thanks for the pointers @igamigo

PR #36, opened 2025-07-31
chore: update arg name to match docs

mmagician:
Alternatively we could update `README.md`, but I think `--output-path` is clearer than `output`


TomasArrachea (approved):
Thanks!

PR #37, opened 2025-07-31
chore: update README to remove account rollbacks

TomasArrachea:
This was removed when integrating the client.


TomasArrachea:
> Might be nice to explain more of how the client is integrated but this can probably be done more comprehensively separately

We can probably tackle this in: https://github.com/0xMiden/miden-faucet/issues/18

PR #40, opened 2025-08-07
chore: add faucet usage docs

TomasArrachea:
Closes https://github.com/0xMiden/miden-faucet/issues/38.

This PR expands the faucet README to explain how to use the API from a programmatic perspective, rather than interacting via the frontend. Also adds a runnable example under `bin/faucet/examples/request_tokens.rs`.

I guess some of this should be moved to `docs/` when we introduce mdbook (https://github.com/0xMiden/miden-faucet/issues/15)

### DIFF
--- a/bin/faucet/README.md
+++ b/bin/faucet/README.md
@@ -127,7 +127,9 @@ const txId = json.tx_id;
 const explorerUrl = json.explorer_url;
 ```
 
-4. **Request note** to download generated private notes
+4. **Request note** to download generated notes
+
+You must complete this step to retrieve private notes. For public notes, normal client sync is sufficient.
 
 ```typescript
 const response = await fetch(`${baseUrl}/get_note?note_id=${noteId}`);
@@ -168,7 +170,7 @@ Check out the complete working examples below. Make sure the faucet is running a
   - `asset_amount` (number, required): Amount of tokens to request
   - `challenge` (string, required): The encoded challenge from the `/pow` endpoint
   - `nonce` (number, required): The nonce used to solve the challenge
-  - `api_key` (string, required): API key for authentication
+  - `api_key` (string, optional): API key for authentication
 - **Response**: JSON object containing:
   - `tx_id` (string): ID of the created transaction
   - `note_id` (string): ID of the created note
--- a/bin/faucet/examples/typescript/request_tokens.ts
+++ b/bin/faucet/examples/typescript/request_tokens.ts
@@ -1,7 +1,7 @@
 import { sha3_256 } from '@noble/hashes/sha3';
 import fs from 'fs';
 
-async function sendPowRequest(baseUrl: string, accountId: string) {
+async function sendPowRequest(baseUrl: string, accountId: string): Promise<{ challenge: string, target: bigint }> {
     const powUrl = new URL('/pow', baseUrl);
     powUrl.searchParams.set('account_id', accountId);
     const powResp = await fetch(powUrl);
@@ -12,7 +12,7 @@ async function sendPowRequest(baseUrl: string, accountId: string) {
     return { challenge, target };
 }
 
-async function solveChallenge(challenge: string, target: bigint) {
+async function solveChallenge(challenge: string, target: bigint): Promise<number> {
     let nonce = 0;
     while (true) {
         nonce = Math.floor(Math.random() * Number.MAX_SAFE_INTEGER);
@@ -45,7 +45,7 @@ async function solveChallenge(challenge: string, target: bigint) {
 }
 
 
-async function getTokens(baseUrl: string, account_id: string, nonce: number, challenge: string) {
+async function getTokens(baseUrl: string, account_id: string, nonce: number, challenge: string): Promise<{ noteId: string, txId: string, explorerUrl: string }> {
     const params = new URLSearchParams({
         account_id: account_id,
         is_private_note: 'true',
@@ -65,7 +65,7 @@ async function getTokens(baseUrl: string, account_id: string, nonce: number, cha
     return { noteId, txId, explorerUrl };
 }
 
-async function downloadNote(baseUrl: string, noteId: string) {
+async function downloadNote(baseUrl: string, noteId: string): Promise<void> {
     const url = `${baseUrl}/get_note?note_id=${noteId}`;
     const response = await fetch(url);
     if (!response.ok) throw new Error(`Get note error: ${response.status} ${await response.text()}`);
@@ -79,7 +79,7 @@ async function downloadNote(baseUrl: string, noteId: string) {
     fs.writeFileSync('note.mno', noteData);
 }
 
-async function main() {
+async function main(): Promise<void> {
     const baseUrl = 'http://localhost:8080';
     const accountId = 'mlcl1qq8mcy8pdvl0cgqfkjzf8efjjsnlzf7q';
 
--- a/bin/faucet/examples/typescript/package.json
+++ b/bin/faucet/examples/typescript/package.json
@@ -9,5 +9,6 @@
   "dependencies": {
     "typescript": "^5.9.2",
     "@noble/hashes": "^1.8.0"
-  }
+  },
+  "type": "module"
 }


bin/faucet/README.md
```diff
@@ -46,5 +46,138 @@ The faucet implements several security measures to prevent abuse:
...
+1. **Request a Proof-of-Work challenge** from `/pow`
+
+```rust
+let base_url = "http://localhost:8080";
+let url = format!("{base_url}/pow?account_id=mlcl1qq8mcy8pdvl0cgqfkjzf8efjjsnlzf7q");
+let response = reqwest::get(&url).await?.error_for_status()?.text().await?;
+let json: serde_json::Value = serde_json::from_str(&response)?;
+let challenge: &str = json["challenge"].as_str().unwrap();
+let target: u64 = json["target"].as_u64().unwrap();
+```
```
bobbinth, line 66: Is it possible to use TypeScript or JavaScript examples? My thinking for this is that the target user here is probably a web developer, and I'm assuming we already have such examples because we do this in the faucet front-end.

We can still have a separate Rust example file (i.e., the `bin/faucet/examples/request_tokens.rs`) but here it may make sense to show to to do these from the web.

bin/faucet/README.md
```diff
@@ -46,5 +46,138 @@ The faucet implements several security measures to prevent abuse:
...
+
+3. **Request tokens** along with your solved challenge to `/get_tokens`
+
+```rust
+let base_url = "http://localhost:8080";
+let params = format!("account_id=mlcl1qq8mcy8pdvl0cgqfkjzf8efjjsnlzf7q\
+    &is_private_note=true\
+    &asset_amount=100\
+    &challenge={}\
+    &nonce={}", challenge, &nonce.to_string());
```
bobbinth, line 103: Not for this PR, but I wonder if we should make the amount a part of the challenge. We could even make the PoW depend on the amount (the more tokens you request the higher PoW). Let's create an issue for this.

bin/faucet/README.md
```diff
@@ -46,5 +46,138 @@ The faucet implements several security measures to prevent abuse:
...
+  - `note_id` (string, required): The ID of the note to retrieve
+- **Response**: JSON object containing:
+  - `note_id` (string): The ID of the requested note
+  - `data_base64` (string): The note data encoded in base64 format. This data should be decoded and saved as a file with `.mno` extension and `application/octet-stream` media type
+
+#### Error Handling
+
+The API returns appropriate HTTP status codes:
+- `200`: Success
+- `400`: Bad request (invalid parameters or rate limited)
```
bobbinth, line 177: Not for this PR, but should we differentiate between invalid parameters and rate limited somehow?

But also, does rate limiting still apply? (i.e., under which circumstances would a request be rate-limited?)

bin/faucet/README.md
```diff
@@ -46,5 +46,138 @@ The faucet implements several security measures to prevent abuse:
...
+  - `account_id` (string, required): The account ID requesting tokens
+  - `is_private_note` (boolean, required): Whether to create a private note
+  - `asset_amount` (number, required): Amount of tokens to request
+  - `challenge` (string, required): The encoded challenge from the `/pow` endpoint
+  - `nonce` (number, required): The nonce used to solve the challenge
+  - `api_key` (string, required): API key for authentication
+- **Response**: JSON object containing:
+  - `tx_id` (string): ID of the created transaction
+  - `note_id` (string): ID of the created note
+  - `explorer_url` (string): URL to view the transaction in the explorer. Only present if available for the current network.
```
bobbinth, line 162: What does "Only present if available for the current network."? How does the faucet know if there is a block explorer?

bin/faucet/README.md
```diff
@@ -46,5 +46,138 @@ The faucet implements several security measures to prevent abuse:
...
+
+#### API Endpoints Reference
+
+**GET /pow**
+- **Purpose**: Request a proof-of-work challenge
+- **Query Parameters**:
+  - `account_id` (string, required): The account ID requesting the challenge
+  - `api_key` (string, optional): API key for authentication
+- **Response**: JSON object containing:
+  - `challenge` (string): The encoded challenge string in hexadecimal format. This string serializes the target, timestamp, account ID, API key, and signature
```
bobbinth, line 146: I'd probably delete the second sentence from here. I don't think we need to explain what goes into the challenge.

bin/faucet/README.md
```diff
@@ -46,5 +46,138 @@ The faucet implements several security measures to prevent abuse:
...
+  - `target` (number): The target value for the proof-of-work challenge. A solution is valid if the hash `H(challenge, nonce)` is less than this target
```
bobbinth, line 147: I would add that for hashing we use `SHA3-256`. btw, is there a way to switch to SHA256 or BLAKE3? (I think I asked this before but don't remember what the answer was).

bobbinth (commented):
Looks good! Thank you! I left some comments inline. The main one is about using TS/JS examples in the README instead of Rust.

Once we have the mdBook, we can probably have articles for both Rust and TS/JS.

bin/faucet/README.md
TomasArrachea: Good idea, created https://github.com/0xMiden/miden-faucet/issues/42

bin/faucet/README.md
TomasArrachea: Yeah I noticed the same when writing this, we could use 429 status code for the rate limited error.

Once the user submits a PoW challenge, it is stored in the cache for the challenge lifetime (default is 30s). During that time, the account cannot submit another challenge so it's "rate limited'.

TomasArrachea: https://github.com/0xMiden/miden-faucet/issues/43

bin/faucet/README.md
TomasArrachea: The faucet tracks the network id where it's running so we use that to match against a list of hardcoded explorer urls. For now it only returns an explorer for testnet:
https://github.com/0xMiden/miden-faucet/blob/9b0d694c7077965148f1aa81a0eba1f910f62eda/bin/faucet/src/network.rs#L52-L57

bin/faucet/README.md
TomasArrachea: Was it this comment maybe? https://github.com/0xMiden/miden-node/pull/831#discussion_r2083438781

should we create an issue for switching to BLAKE3 or make it configurable?

bin/faucet/README.md
```diff
@@ -143,8 +143,8 @@ cargo run --example request_tokens
   - `account_id` (string, required): The account ID requesting the challenge
   - `api_key` (string, optional): API key for authentication
 - **Response**: JSON object containing:
-  - `challenge` (string): The encoded challenge string in hexadecimal format. This string serializes the target, timestamp, account ID, API key, and signature
-  - `target` (number): The target value for the proof-of-work challenge. A solution is valid if the hash `H(challenge, nonce)` is less than this target
+  - `challenge` (string): The encoded challenge string in hexadecimal format
+  - `target` (number): The target value for the proof-of-work challenge. A solution is valid if the hash `H(challenge, nonce)` is less than this target. We use `SHA-256` for this
```
bobbinth, line 147: I thought we are using `SHA3-256`, or are we using `SHA256`?

I would also say "As the hash function we use ...".

bin/faucet/examples/typescript/request_tokens.ts
```diff
@@ -0,0 +1,95 @@
+import { sha3_256 } from '@noble/hashes/sha3';
+import fs from 'fs';
+
+async function sendPowRequest(baseUrl: string, accountId: string) {
```
bobbinth, line 4: For this and other functions, could we specify the return type? (otherwise, it is not immediately clear which functions return values and which ones do not).

bin/faucet/examples/typescript/request_tokens.ts
```diff
@@ -0,0 +1,95 @@
+import { sha3_256 } from '@noble/hashes/sha3';
```
bobbinth, line 1: Not for this PR, but maybe we should use "built-in" hash function libraries via the WebCrypto APIs.

bin/faucet/README.md
```diff
@@ -46,5 +46,151 @@ The faucet implements several security measures to prevent abuse:
...
+if (!response.ok) throw new Error(`Get tokens error: ${response.status} ${await response.text()}`);
+
+const text = await response.text();
+const json = JSON.parse(text);
+const noteId = json.note_id;
+const txId = json.tx_id;
+const explorerUrl = json.explorer_url;
+```
+
+4. **Request note** to download generated private notes
```
bobbinth, line 130: This would work for public notes as well, right? If so, maybe I'd phrase it a bit differently.

bin/faucet/README.md
```diff
@@ -46,5 +46,151 @@ The faucet implements several security measures to prevent abuse:
...
+
+**GET /get_tokens**
+- **Purpose**: Request tokens
+- **Query Parameters**:
+  - `account_id` (string, required): The account ID requesting tokens
+  - `is_private_note` (boolean, required): Whether to create a private note
+  - `asset_amount` (number, required): Amount of tokens to request
+  - `challenge` (string, required): The encoded challenge from the `/pow` endpoint
+  - `nonce` (number, required): The nonce used to solve the challenge
+  - `api_key` (string, required): API key for authentication
```
bobbinth, line 171: Is `api_key` required? I thought it would be optional, and if not submitted, and empty string would be assumed.

bin/faucet/README.md
```diff
@@ -46,5 +46,151 @@ The faucet implements several security measures to prevent abuse:
...
+  - `challenge` (string): The encoded challenge string in hexadecimal format
+  - `target` (number): The target value for the proof-of-work challenge. A solution is valid if the hash `H(challenge, nonce)` is less than this target. As the hashing function we use `SHA3-256`
+  - `timestamp` (number): The timestamp when the challenge was created (seconds since UNIX epoch)
+
+**GET /get_tokens**
+- **Purpose**: Request tokens
+- **Query Parameters**:
+  - `account_id` (string, required): The account ID requesting tokens
+  - `is_private_note` (boolean, required): Whether to create a private note
+  - `asset_amount` (number, required): Amount of tokens to request
```
bobbinth, line 168: I would clarify that this is in base units. Also, I believe we currently support only some subset of amounts, right? I think at the API level we should just impose an upper limit (e.g., less than 1000 tokens) - but this would be for another PR.

bobbinth (approved):
Looks good! Thank you! I left a few small comments inline.

bin/faucet/examples/typescript/request_tokens.ts
bobbinth: Yep, if we move to WebCrypto, I think we should switch to SHA256.

bin/faucet/README.md
TomasArrachea: > I believe we currently support only some subset of amounts, right?

Yes, the CLI has a `--asset-amounts` parameter that sets the possible options. Actually these amounts are not in base units.

> at the API level we should just impose an upper limit

Makes sense. Created https://github.com/0xMiden/miden-faucet/issues/46

bin/faucet/examples/typescript/request_tokens.ts
TomasArrachea: Ok, created https://github.com/0xMiden/miden-faucet/issues/47

bin/faucet/README.md
bobbinth: Let's create an issue (unless we already have one) to make sure all internal amounts are in base units. The only time we should show token amounts is for display purposes.

bobbinth (approved):
Looks good! Thank you!

bin/faucet/README.md
TomasArrachea: https://github.com/0xMiden/miden-faucet/issues/49

PR #52, opened 2025-08-12
feat: remove minting amount options at API level

TomasArrachea:
Closes https://github.com/0xMiden/miden-faucet/issues/46, https://github.com/0xMiden/miden-faucet/issues/49

This PR:
- removes the validation that checks if the amount of tokens claimed belongs to one of the minting options. This set of options is now only used on the frontend to display options to the user.
Instead, the endpoint now validates that the requested amount is less than the max claimable amount - this is set by a new CLI parameter.
- refactors the token flow to use `AssetAmounts` to ensure all internal amounts are base units. The frontend now formats the amounts to show it as token amounts.

In https://github.com/0xMiden/miden-faucet/issues/42, we will make the PoW difficulty to scale with the requested amount.


crates/faucet/src/types.rs
```diff
@@ -64,36 +17,45 @@ impl std::fmt::Display for AssetAmount {
...
-                    "Maximum fungible asset value allowed by the network is {}",
-                    AssetAmount::MAX
-                )
-                .as_str(),
-            )
-        })
+    /// Subtracts another [`AssetAmount`] from the current one and returns the result if it is
+    /// valid.
+    pub fn checked_sub(self, other: Self) -> Option<Self> {
+        Self::new(self.0.checked_sub(other.0)?).ok()
```
bobbinth, line 46: This naming is fine but I also don't mind just `add()` and `sub()` because we don't have the unchecked versions.

bobbinth (approved):
Looks good! Thank you! We can probably refactor it a bit more - but that can be done in future PRs.

PR #54, opened 2025-08-18
feat: use `ServeDir` for frontend

TomasArrachea:
This PR changes how the frontend is served to avoid having a specific function for each file. Instead, we use `axum::ServeDir` to serve the whole `/frontend` directory.


bobbinth:
> Just my 2c - we previously moved away from `ServeDir` while this was in the node because there was no compile time checking on which files exist at which routes and where, and with which metadata.
> 
> As an example, I think right now if you type in a bad resource path you get `index.html` as the response instead of a `404`.

I like the simplicity of serving just from the directory - but it would be nice if we returned a 404 on non-existent URL path. @TomasArrachea - would this be difficult to do?

In the longer term, we were thinking of moving the API part to a different sub-domain - e.g., `api.faucet.testnet.miden.io`.

bobbinth:
> I like the simplicity of serving just from the directory - but it would be nice if we returned a 404 on non-existent URL path. @TomasArrachea - would this be difficult to do?

I'll merge as is so that we can deploy the faucet - but seems like we can return 404 with a custom page relatively easily (see [here](https://docs.rs/tower-http/latest/tower_http/services/struct.ServeDir.html#method.not_found_service)). @TomasArrachea - could you implement this in a follow-up PR?

TomasArrachea:
Yes! Adding a custom 404 on any non-existent path should simple. I'll make the follow up PR.

PR #56, opened 2025-08-19
feat: compile transaction script once

### LINKED ISSUE #3: Compile faucet transaction script only once
With https://github.com/0xMiden/miden-base/issues/1262, we can now address [this TODO](https://github.com/0xMiden/miden-node/blob/1ca60377534d7afd8c5bc62acbf99738771365c3/bin/faucet/src/faucet/store.rs#L53-L58) by using transaction args to pass the targets of the minting. 
We'd need to compile the transaction script once and modify it to receive the target account IDs and requested note type (and potentially note tag as well? though it can be derived from the target account ID) from the transaction args. 

Benefits would basically be working with a single precompiled transaction script as opposed to compiling each time, which in turn implies not indefinitely growing the `MastForestStore`. However, because the faucet is currently using the `AccountInterface` to build the transaction script, this might imply a bigger refactor than it would look like initially and so maybe the benefits don't outweigh the effort too much (since the faucet is performing well overall).

TomasArrachea:
Closes #3 

This PR changes how transactions are built. Instead of using the standard transaction script for note creation, it uses a custom script that receives the note data through the advice map, and accesses it from the script to generate the p2id notes.


crates/faucet/src/scripts/transaction_script.masm
bobbinth, line 1: In `miden-base`, we usually put assembly files into a separate `asm` directory. Here it could be `crates/faucet/asm/tx_scripts/mint.masm` (or something like that).

We also usually assemble the scripts at build time (via the `build.rs` file).

crates/faucet/Cargo.toml
```diff
@@ -16,7 +16,8 @@ workspace = true
 
 [dependencies]
 # Miden dependencies.
-miden-client = { features = ["sqlite", "tonic"], workspace = true }
+miden-client  = { features = ["sqlite", "tonic"], workspace = true }
+miden-objects = { branch = "next", default-features = false, git = "https://github.com/0xMiden/miden-base" }
```
bobbinth, line 20: Why do we need to import `miden-objects` here?

igamigo: AFAIK it's for using `Rpo256::hash_elements`. It has since been [re-exported from the client](https://github.com/0xMiden/miden-client/pull/1175), so this should get updated here soon.

TomasArrachea: Yes, I removed it in [73cfcfd](https://github.com/0xMiden/miden-faucet/pull/56/commits/73cfcfd3b41613e639908a84da5298d61392c6c3) to import `Rpo256` from miden-client.

fkrause98 (approved):
Looks good

crates/faucet/src/lib.rs
```diff
@@ -230,12 +236,32 @@ impl Faucet {
...
+        let n = notes.len() as u64;
+        let mut note_data = vec![Felt::new(n)];
+        for note in notes {
+            // SAFETY: these are p2id notes with only one fungible asset
+            let amount = note.assets().iter().next().unwrap().unwrap_fungible().amount();
+
+            note_data.push(note.recipient().digest()[0]);
+            note_data.push(note.recipient().digest()[1]);
+            note_data.push(note.recipient().digest()[2]);
+            note_data.push(note.recipient().digest()[3]);
```
bobbinth, line 251: Would something like `note_data.extend(recipient().digest().iter())` not work here?

crates/faucet/src/lib.rs
```diff
@@ -230,12 +236,32 @@ impl Faucet {
...
+            note_data.push(Felt::from(note.metadata().note_type()));
+            note_data.push(Felt::from(note.metadata().tag()));
+            note_data.push(Felt::new(amount));
+        }
+        let note_data_commitment = Rpo256::hash_elements(&note_data);
+        let advice_map = [(note_data_commitment, note_data)];
+
+        let tx = TransactionRequestBuilder::new()
```
bobbinth, line 259: nit: I would probably rename `tx` to `tx_request`

bobbinth (approved):
Looks good! Thank you!

In the longer run, we should put the transaction script MASM into a separate crate and make it appear as a "package". This is how eventually the compiler will be used to create transaction scripts in Rust (i.e., there will be a separate crate per transaction script).

The same crate may contain some kind of "constructor" that could make building transactions using this script easier.

Let's create an issue for these.

crates/faucet/build.rs
```diff
@@ -0,0 +1,92 @@
...
+        let script = TransactionScript::new(code);
+
+        let bytes = script.to_bytes();
+
+        let masm_file_name = masm_file_path.file_name().expect("file name should exist");
+        let mut txs_file_path = target_dir.join(masm_file_name);
+
+        // write the binary TXS to the output dir
+        txs_file_path.set_extension("txs");
+        fs::write(txs_file_path, bytes).expect("should write .txs file");
```
mmagician, line 54: nit: In miden-base we have similar logic, and there we suffix our files `.masb`. For reasons of consistency, I'd stick to one convention: https://github.com/0xMiden/miden-base/blob/643dd211eee1302d5a92ddf66b0402d9dcbc5c0e/crates/miden-lib/build.rs#L344

crates/faucet/build.rs
```diff
@@ -0,0 +1,92 @@
...
+    let entries = fs::read_dir(dir_path)?;
+    for entry in entries {
+        let file_path = entry?.path();
+        if is_masm_file(&file_path)? {
+            files.push(file_path);
+        }
+    }
+
+    Ok(files)
+}
```
mmagician, line 76: Perhaps a broader point (not for this PR) is whether we could re-use that code from miden-lib. Even there we have two places where `get_masm_files` is defined: in `build.rs` and in `src/transaction/mod.rs`.
Sadly we can't re-use the code easily because `build.rs` can't depend on code defined in its own (yet un-built) crate.
An interim solution would be to extract the definition from `src/transaction/mod.rs` to say, `src/utils/', and we could reuse it both in `src/transaction/mod.rs` and here (and `build.rs` keeps its own definition).
Another solution would be to introduce a helper crate where all three locations could pull `get_masm_files` from - but that's an overkill unless we have other uses for this helper crate (and also it's not too much code that is duplicated).

mmagician (approved):
LGTM ✅

crates/faucet/build.rs
bobbinth: I think one subtle difference here is that this is a direct serialization of `TransactionScript` struct, while in `miden-base` we serialize a `Program` struct which, after deserialization, gets converted into a `NoteScript` struct. So, we either should:

- Serialize `Program` with `.masb` extension, or
- Serialize `TransactionScript` with some other extension (i.e., `.txs`).

I'm fine with either approach tbh.

crates/faucet/build.rs
bobbinth: Agreed that it would be good to encapsulate this logic a bit better - but I'm viewing this as a temporary solution on our way to figure out out how to structure such projects properly. The longer-term solution would be to put the script itself into a separate crate (and wrap it into a "package") to align the methodology with how the compiler will generate package.

crates/faucet/build.rs
bobbinth: Ah yes - I remember now, the first approach requires the user to know about the `Program` struct. So, I probably have a slight preference toward the `.txs` approach.

It may be good to have standard serialization defined for transaction and note scripts in `miden-base` (i.e., to/from file method) - but that's for another time.

PR #63, opened 2025-08-25
feat: make `explorer_url` configurable and remove `network` cli param

TomasArrachea:
This PR adapts the faucet to the latest `miden-base` breaking changes.

Additionally, replaced the hardcoded `explorer_url` with a CLI parameter (closes https://github.com/0xMiden/miden-faucet/issues/48). The `network` parameter was removed, now the network is interpreted directly from the node url.


bin/faucet/.env
```diff
@@ -15,3 +15,4 @@ MIDEN_FAUCET_POW_BASELINE=
 MIDEN_FAUCET_API_KEYS=
 MIDEN_FAUCET_ENABLE_OTEL=false
 MIDEN_FAUCET_STORE=
+MIDEN_FAUCET_EXPLORER_URL=https://testnet.midenscan.com/
```
igamigo, line 18: I think I'd leave this with no default, otherwise it seems like it would be very easy to have an incorrect setting

bin/faucet/frontend/index.js
```diff
@@ -236,9 +238,9 @@ class MidenFaucet {
             completedPublicModal.classList.add('active');
 
             const explorerButton = document.getElementById('explorer-button');
-            if (mintingData.explorer_url) {
+            if (this.explorer_url) {
                 explorerButton.style.display = 'block';
-                explorerButton.onclick = () => window.open(mintingData.explorer_url + '/tx/' + mintingData.tx_id, '_blank');
+                explorerButton.onclick = () => window.open(this.explorer_url + '/tx/' + mintingData.tx_id, '_blank');
```
igamigo, line 243: Does this get correctly concatenated? The default URL has a trailing `/` and then you are concatenating against a `/tx` (although I don't know how a `Url` gets serialized).

bin/faucet/src/api/get_metadata.rs
```diff
@@ -21,11 +23,12 @@ impl Serialize for Metadata {
...
     {
         use serde::ser::SerializeStruct;
-        let mut state = serializer.serialize_struct("Metadata", 4)?;
-        state.serialize_field("id", &self.id.to_bech32())?;
+        let mut state = serializer.serialize_struct("Metadata", 5)?;
+        state.serialize_field("id", &self.id.inner().to_hex())?;
         state.serialize_field("issuance", &self.issuance.read().base_units())?;
         state.serialize_field("max_supply", &self.max_supply.base_units())?;
         state.serialize_field("decimals", &self.decimals)?;
+        state.serialize_field("explorer_url", &self.explorer_url)?;
```
igamigo, line 31: nit: This and the account ID would be static fields, so I wonder if there is a way to somehow embed them in the HTML/JS responses instead of getting them dynamically.

igamigo (approved):
Left a few minor comments but LGTM!

bobbinth:
With `miden-base` v0.11.1 we can now convert any account ID to an address (using the `AddressInterface::Unspecified` variant for the interface). Let's use this here for now.

In the future we may still switch back to using account ID - but for this release we could keep it as an address.

TomasArrachea:
> With miden-base v0.11.1 we can now convert any account ID to an address (using the AddressInterface::Unspecified variant for the interface). Let's use this here for now.

Right, sorry, this change was in a separate PR. I merged it here so now its updated to miden-base `0.11.1`.

This PR addresses both the issue of removing the `explorer_url` and updating the miden versions of miden-client, miden-base, miden-node. It probably should have been two separate PRs but it got mixed. 

I tested this agains the devnet node (running on v0.11) and it's working correctly.

After the miden-client release is ready, I'll make a separate PR to use it.

Cargo.lock
TomasArrachea, line 1: I also ran `cargo update` so that explains the big diff in the cargo.lock

bin/faucet/src/api/mod.rs
```diff
@@ -201,3 +203,17 @@ impl FromRef<Server> for PoW {
         input.pow.clone()
     }
 }
+
+// ERRORS
+// ================================================================================================
+
+/// Errors that can occur when parsing an account ID or address.
+#[derive(Debug, thiserror::Error)]
+pub enum AccountError {
```
bobbinth, line 212: Maybe this should be named `AddressError` or `RecipientError`?

bin/faucet/src/api/get_tokens.rs
```diff
@@ -197,11 +194,16 @@ impl RawMintRequest {
...
-            AccountId::from_bech32(&self.account_id).map(|(_, account_id)| account_id)
+            Address::from_bech32(&self.account_id)
+                .map_err(AccountError::ParseAddress)
+                .and_then(|(_, address)| match address {
+                    Address::AccountId(account_id_address) => Ok(account_id_address.id()),
+                    _ => Err(AccountError::AddressNotIdBased),
+                })
         }
-        .map_err(MintRequestError::AccountId)?;
+        .map_err(MintRequestError::AccountError)?;
```
bobbinth, line 206: This is fine for now, but in the future we should make this more generic. Basically, based on address/id we should be able to figure out how to build notes. Specifics of this are still TBD and maybe a part of this should live in the client (or maybe in `miden-base`) - but let's create an issue to figure out how to build notes based on addresses.

bin/faucet/.env
bobbinth, line 1: Since the set of environment variables changed, we may need to make some updates in deployment scripts in the infra repo. cc @sergerad.

crates/faucet/src/lib.rs
```diff
@@ -144,7 +152,7 @@ impl Faucet {
             Some(url) => Arc::new(RemoteTransactionProver::new(url)),
             None => Arc::new(LocalTransactionProver::default()),
         };
-        let id = FaucetId::new(account.id(), network_id);
+        let id = FaucetId::new(account.id(), endpoint.to_network_id());
```
bobbinth, line 155: I think this is fine for now, but mapping endpoint to a network ID may be fragile (e.g., in theory just an IP address should work as an endpoint). So, we may need to specify the network ID regardless.

Let's create an issue for this.

bobbinth (approved):
Looks good! Thank you! I left some comments inline - but they are all for the future/follow-ups.

PR #64, opened 2025-08-25
feat: update default `--max-claimable-amount`

TomasArrachea:
Context: https://github.com/0xMiden/miden-faucet/issues/62

This PR sets the default value for `--max-claimable-amount` at 1000000000. This is 1000 tokens when using 6 decimals, the value used for the testnet faucet.


sergerad (approved):

igamigo (approved):
LGTM!

PR #66, opened 2025-08-26
feat: show explorer button only when available

TomasArrachea:
This PR changes the frontend to show the explorer button only when available, i.e. only on testnet.

Context: https://github.com/0xMiden/miden-faucet/issues/65


bobbinth (approved):
Looks good! Thank you!

PR #67, opened 2025-08-27
feat: add `miden-faucet-pow` crate

TomasArrachea:


crates/pow/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+authors.workspace      = true
+description            = "Miden faucet's PoW"
+edition.workspace      = true
+homepage.workspace     = true
+keywords               = ["faucet", "miden", "pow"]
+license.workspace      = true
+name                   = "miden-faucet-pow"
```
bobbinth, line 8: I would probably make this crate a bit more generic and we could name it `miden-pow-rate-limiter` or something like that.

crates/pow/src/lib.rs
```diff
@@ -51,14 +66,14 @@ impl PoW {
     }
 
     /// Generates a new challenge.
-    pub fn build_challenge(&self, request: PowRequest) -> Challenge {
+    pub fn build_challenge(&self, account_id: AccountId, api_key: ApiKey) -> Challenge {
```
bobbinth, line 69: Related to the above comment: could we make this more generic (i.e., not tied to account IDs or even API keys). For example, this could be something like:
```rust
pub fn build_challenge(&self, 
    requestor: impl Into<[u8; 32]>,
    domain: impl Into<[u8; 32]>,
) -> Challenge {
    ...
}
```
And then internally we just work with these generic notions of "requestor" and "domain". I haven't thought it through completely though - so, not sure if it would cause some unforeseen side effects.

crates/pow/src/lib.rs
```diff
@@ -2,32 +2,51 @@ use std::num::NonZeroUsize;
...
+
 // POW
 // ================================================================================================
 
+/// Proof-of-Work implementation.
+///
+/// This struct is used to generate and validate `PoW` challenges.
 #[derive(Clone)]
-pub(crate) struct PoW {
+pub struct PoW {
```
bobbinth, line 21: I wonder if we should rename this struct (and update the associated docs) into `PoWRateLimiter` (or maybe just `RateLimiter`).

crates/pow/src/lib.rs
```diff
@@ -2,32 +2,51 @@ use std::num::NonZeroUsize;
...
+    /// The server secret used to sign and validate challenges.
     secret: [u8; 32],
+    /// The cache used to store submitted challenges.
     challenge_cache: Arc<Mutex<ChallengeCache>>,
```
bobbinth, line 25: nit: let's rename this into `challenges`.

crates/pow/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+repository.workspace   = true
+rust-version.workspace = true
+version.workspace      = true
+
+[lints]
+workspace = true
+
+[dependencies]
+# Miden dependencies.
+miden-client = { workspace = true }
```
bobbinth, line 19: What would we need to do to get rid of the `miden-client` dependency from here?

TomasArrachea: It's only being used for:
```rust
use miden_client::utils::{ToHex, hex_to_bytes};
```
We should definitely remove it. Should we move the `hex_to_bytes` util to `miden-formatting`? `ToHex` is already there, so we can just use that dependency. Otherwise we can just simply add our own implementation for these.

PR #68, opened 2025-08-27
feat: add requested amount to PoW challenge

TomasArrachea:
closes https://github.com/0xMiden/miden-faucet/issues/42.

This PR adds a new `amount` attribute to the `Challenge`. It is used to compute the difficulty of the challenges. Now, the target is also divided by `amount / pow_base_difficulty_amount`.

If the requested amount is lower or equal to the set `pow_base_difficulty_amount`, the difficulty stays the same. But if a higher amount is requested, the difficulty will increase.

The resulting formula for the PoW target is:
```
target = max_target / (load_difficulty * request_complexity)
max_target = u64::MAX >> baseline
load_difficulty = (num_active_challenges / challenges_per_difficulty) + 1
request_complexity = (amount / base_difficulty_amount) + 1
```

### DIFF
--- a/bin/faucet/src/api/get_tokens.rs
+++ b/bin/faucet/src/api/get_tokens.rs
@@ -9,7 +9,7 @@ use miden_pow_rate_limiter::ChallengeError;
 use serde::{Deserialize, Serialize};
 use tokio::sync::mpsc::error::TrySendError;
 use tokio::sync::oneshot;
-use tracing::instrument;
+use tracing::{Instrument, info_span, instrument};
 
 use crate::COMPONENT;
 use crate::api::{AccountError, Server};
@@ -20,7 +20,7 @@ use crate::error_report::ErrorReport;
 // ================================================================================================
 
 #[instrument(
-    parent = None, target = COMPONENT, name = "faucet.server.get_tokens", skip_all,
+    parent = None, target = COMPONENT, name = "server.get_tokens", skip_all, err,
     fields(
         account_id = %request.account_id,
         is_private_note = %request.is_private_note,
@@ -34,23 +34,22 @@ pub async fn get_tokens(
     let (mint_response_sender, mint_response_receiver) = oneshot::channel();
 
     let validated_request = request.validate(&server).map_err(GetTokenError::InvalidRequest)?;
-    let requested_amount = validated_request.asset_amount.base_units();
 
-    let span = tracing::Span::current();
-    span.record("account", validated_request.account_id.to_hex());
-    span.record("amount", requested_amount);
-    span.record("note_type", validated_request.note_type.to_string());
-
-    server
-        .mint_state
-        .request_sender
-        .try_send((validated_request, mint_response_sender))
-        .map_err(|err| match err {
-            TrySendError::Full(_) => GetTokenError::FaucetOverloaded,
-            TrySendError::Closed(_) => GetTokenError::FaucetClosed,
-        })?;
+    let enqueue_span = info_span!(target: COMPONENT, "server.get_tokens.enqueue");
+    {
+        let _enter = enqueue_span.enter();
+        server
+            .mint_state
+            .request_sender
+            .try_send((validated_request, mint_response_sender))
+            .map_err(|err| match err {
+                TrySendError::Full(_) => GetTokenError::FaucetOverloaded,
+                TrySendError::Closed(_) => GetTokenError::FaucetClosed,
+            })?;
+    }
 
     let mint_response = mint_response_receiver
+        .instrument(info_span!(target: COMPONENT, "server.get_tokens.await_mint"))
         .await
         .map_err(|_| GetTokenError::FaucetReturnChannelClosed)?
         .map_err(GetTokenError::MintError)?;
@@ -61,7 +60,7 @@ pub async fn get_tokens(
     }))
 }
 
-#[derive(Serialize)]
+#[derive(Serialize, Debug)]
 pub struct GetTokensResponse {
     tx_id: String,
     note_id: String,
@@ -114,11 +113,17 @@ pub enum MintRequestError {
     MissingPowParameters,
 }
 
+#[derive(Debug, thiserror::Error)]
 pub enum GetTokenError {
-    InvalidRequest(MintRequestError),
-    MintError(MintError),
+    #[error("invalid request: {0}")]
+    InvalidRequest(#[source] MintRequestError),
+    #[error("mint error: {0}")]
+    MintError(#[source] MintError),
+    #[error("faucet overloaded")]
     FaucetOverloaded,
+    #[error("faucet closed")]
     FaucetClosed,
+    #[error("faucet return channel closed")]
     FaucetReturnChannelClosed,
 }
 
@@ -198,7 +203,6 @@ impl RawMintRequest {
     ///   - the nonce is missing or doesn't solve the challenge
     ///   - the challenge timestamp is expired
     ///   - the challenge has already been used
-    #[instrument(level = "debug", target = COMPONENT, name = "faucet.server.validate", skip_all)]
     fn validate(self, server: &Server) -> Result<MintRequest, MintRequestError> {
         let note_type = if self.is_private_note {
             NoteType::Private
--- a/bin/faucet/src/main.rs
+++ b/bin/faucet/src/main.rs
@@ -210,7 +210,8 @@ async fn main() -> anyhow::Result<()> {
     let cli = Cli::parse();
 
     // Configure tracing with optional OpenTelemetry exporting support.
-    logging::setup_tracing(cli.command.open_telemetry()).context("failed to initialize logging")?;
+    let _otel_guard = logging::setup_tracing(cli.command.open_telemetry())
+        .context("failed to initialize logging")?;
 
     Box::pin(run_faucet_command(cli)).await
 }
--- a/bin/faucet/src/api/get_metadata.rs
+++ b/bin/faucet/src/api/get_metadata.rs
@@ -6,7 +6,10 @@ use miden_client::utils::RwLock;
 use miden_faucet_lib::FaucetId;
 use miden_faucet_lib::types::AssetAmount;
 use serde::{Serialize, Serializer};
+use tracing::instrument;
 use url::Url;
+
+use crate::COMPONENT;
 
 /// Describes the faucet metadata needed to show on the frontend.
 pub struct Metadata {
@@ -38,6 +41,7 @@ impl Serialize for Metadata {
 // ENDPOINT
 // ================================================================================================
 
+#[instrument(parent = None, target = COMPONENT, name = "server.get_metadata", skip_all)]
 pub async fn get_metadata(State(metadata): State<&'static Metadata>) -> Json<&'static Metadata> {
     Json(metadata)
 }
--- a/bin/faucet/src/api/get_note.rs
+++ b/bin/faucet/src/api/get_note.rs
@@ -8,7 +8,7 @@ use miden_client::note::NoteId;
 use miden_client::store::{NoteExportType, NoteFilter};
 use miden_client::utils::Serializable;
 use serde::Deserialize;
-use tracing::instrument;
+use tracing::{Instrument, info_span, instrument};
 
 use crate::COMPONENT;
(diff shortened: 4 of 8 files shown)


bin/faucet/src/main.rs
```diff
@@ -89,9 +89,11 @@ pub enum Command {
         #[arg(long = "account", value_name = "FILE", env = ENV_ACCOUNT_PATH)]
         faucet_account_path: PathBuf,
 
-        /// The maximum amount of assets base units that can be dispersed on each request.
-        #[arg(long = "max-claimable-amount", value_name = "U64", env = ENV_MAX_CLAIMABLE_AMOUNT, default_value = "1000000000")]
-        max_claimable_amount: u64,
+        /// The maximum amount of assets base units that can be dispersed on each request. It is
```
mmagician, line 92:
```suggestion
        /// The maximum amount of asset's base units that can be dispersed on each request. It is
```

mmagician (approved):
LGTM ✅

bobbinth:
> If the requested amount is the maximum claimable, the difficulty stays the same. But if a lower amount is requested, the difficulty will be lower.

As mentioned in a different comment, could this potentially lead to an issue where someone requesting 1 token could perform very little work and thus, be able to spam the network. Generally, would be great to see what PoW values would be in a handful of scenarios (e.g., low load + small amount, low load + large amount, high load + small amount etc.).

bobbinth:
@TomasArrachea - is this ready for another review?

crates/pow/README.md
```diff
@@ -38,12 +38,19 @@ A challenge solution is valid when:
...
-The system automatically adjusts challenge difficulty based on usage:
-- **Target calculation**: `max_target / difficulty`
-- **Max target**: `u64::MAX >> baseline`
-- **Difficulty**: `max(num_active_challenges << growth_rate, 1)`
+The system automatically adjusts challenges difficulty based on usage and the request complexity. 
 
-This means as more users solve challenges, the difficulty increases exponentially, providing automatic rate limiting.
+The challenge target $t$ is computed as follows: $$ t = \dfrac{2^{64}} {d} $$
+
+We call $d$ "difficulty" such that $log_2(d)$ is the number of bits of work that we need to do to solve the challenge. We compute it as: $$ d = 2^b \centerdot c \centerdot n \centerdot g $$
```
bobbinth, line 45: The formulas don't render for me properly. I think they need to be either in single `$` or if we use `$$`, then they need to be in separate code blocks. For example:

```
$$
d = 2^b \centerdot c \centerdot n \centerdot g
$$
```

crates/pow/README.md
```diff
@@ -22,9 +22,9 @@ The `PoWRateLimiterConfig` struct allows you to customize the behavior of the ra
 
 - **`challenge_lifetime`**: How long a challenge remains valid after generation. After this duration, challenges expire and cannot be submitted. Choose based on expected solving time and security requirements.
 
-- **`growth_rate`**: Controls how aggressively difficulty increases with more active challenges. The difficulty formula is `num_active_challenges << growth_rate`. Higher values mean more aggressive rate limiting. Typical values: 1-3.
+- **`growth_rate`**: Controls how aggressively difficulty increases with more active challenges. The number of active challenges gets multiplied by the growth rate to compute the load difficulty, so higher values mean more aggressive rate limiting.
 
-- **`baseline`**: Sets the initial difficulty baseline. The maximum target is calculated as `u64::MAX >> baseline`. Higher baseline values make challenges harder from the start. Range: 0-63.
+- **`baseline`**: Sets the initial difficulty baseline. Higher baseline values make challenges harder from the start. Range: 0-63.
```
bobbinth, line 27: Maybe not for this PR: but I'd limit the range here to 0 - 32. A baseline with 32 bits of difficulty would probably already be out of reach in most practical settings.

crates/pow/src/lib.rs
```diff
@@ -66,42 +63,54 @@ impl PoWRateLimiter {
...
-        let max_target = u64::MAX >> self.config.baseline;
-        let difficulty = usize::max(num_challenges << self.config.growth_rate.get(), 1);
-        max_target / difficulty as u64
+            .num_challenges_for_domain(domain) as u64;
+
+        #[allow(clippy::cast_precision_loss, reason = "num_challenges is smaller than f64::MAX")]
+        #[allow(clippy::cast_sign_loss, reason = "growth_rate and num_challenges are positive")]
+        let load_difficulty = 2_u64.pow(self.config.baseline.into())
+            * ((num_challenges + 1) as f64 * self.config.growth_rate).ceil() as u64;
+        let request_difficulty = load_difficulty * request_complexity;
```
bobbinth, line 112: Let's use safe operations here (or catch potential overflows in some other way). If there is an overflow anywhere, we should set the difficulty to `u64::MAX`.

Also, it may be good to split the `load_difficulty` computation into two steps (to make it a bit easier to follow.

bin/faucet/.env
```diff
@@ -11,6 +11,7 @@ MIDEN_FAUCET_POW_CHALLENGE_LIFETIME=
 MIDEN_FAUCET_POW_CLEANUP_INTERVAL=
 MIDEN_FAUCET_POW_GROWTH_RATE=
 MIDEN_FAUCET_POW_BASELINE=
+MIDEN_FAUCET_POW_BASE_DIFFICULTY_AMOUNT=
```
bobbinth, line 14: I wonder if the name should be just `MIDEN_FAUCET_BASE_AMOUNT`. I know it affects the PoW, but also, we could say that this is a more general setting for a "normal amount" that we expect to be requested.

bobbinth (commented):
Looks good! Thank you. I left some more comments inline.

crates/pow/src/challenge.rs
```diff
@@ -5,14 +5,16 @@ use sha2::{Digest, Sha256};
...
+#[derive(Debug, Clone, PartialEq, PartialOrd)]
 pub struct Challenge {
     /// The target used to validate the challenge solution. A lower target makes the challenge more
     /// difficult to solve. A solution is valid if the hash `H(challenge, nonce)`, interpreted as a
     /// big-endian u64 from the first 8 bytes, is less than this target value.
     pub(crate) target: u64,
     /// The timestamp of the challenge creation.
     pub(crate) timestamp: u64,
+    /// The request complexity, used to scale the difficulty of the challenge.
+    pub(crate) request_complexity: u64,
```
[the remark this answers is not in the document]
TomasArrachea: The target is derived from the complexity but doesn't include it. We need to include it separately to validate that the requested amount is the same as submitted. Meaning, use requests a challenge for a certain token amount and then submits it minting that same amount.
https://github.com/0xMiden/miden-faucet/blob/af38e4ad0ff4740c54dc8b442057dcf0f152c111/crates/pow/src/lib.rs#L149-L152

docs/src/getting-started/cli.md
```diff
@@ -73,15 +74,17 @@ export MIDEN_FAUCET_NODE_URL=https://rpc.testnet.miden.io
...
+export MIDEN_FAUCET_MAX_CLAIMABLE_AMOUNT=1000000000
 
 # Proof of Work
 export MIDEN_FAUCET_POW_SECRET=your-secret-here
 export MIDEN_FAUCET_POW_BASELINE=12
 export MIDEN_FAUCET_POW_CHALLENGE_LIFETIME=30s
-export MIDEN_FAUCET_POW_GROWTH_RATE=1
+export MIDEN_FAUCET_POW_CLEANUP_INTERVAL=2s
+export MIDEN_FAUCET_POW_GROWTH_RATE=0.1
+export MIDEN_FAUCET_POW_BASE_DIFFICULTY_AMOUNT=100000000
```
bobbinth, line 85: This should be `MIDEN_FAUCET_BASE_AMOUNT`, right? Also, I'd move this setting to the "Basic configuration" section.

crates/pow/README.md
```diff
@@ -38,12 +38,25 @@ A challenge solution is valid when:
...
-- **Target calculation**: `max_target / difficulty`
-- **Max target**: `u64::MAX >> baseline`
-- **Difficulty**: `max(num_active_challenges << growth_rate, 1)`
+The system automatically adjusts challenges difficulty based on usage and the request complexity. 
 
-This means as more users solve challenges, the difficulty increases exponentially, providing automatic rate limiting.
+The challenge target $t$ is computed as follows: 
+$$
+t = \dfrac{2^{64}} {d}
+$$
```
bobbinth, line 46: This still doesn't render correctly - I believe there should be a newline before `$$`.

crates/pow/README.md
```diff
@@ -38,12 +38,25 @@ A challenge solution is valid when:
...
+
+We call $d$ "difficulty" such that $log_2(d)$ is the number of bits of work that we need to do to solve the challenge. We compute it as: 
+$$
+d = 2^b \centerdot c \centerdot n \centerdot g
+$$
```
bobbinth, line 51: Same comment as above.

crates/pow/README.md
```diff
@@ -38,12 +38,25 @@ A challenge solution is valid when:
...
+
+Where:
+- $b$ is the baseline difficulty (in bits).
+- $c$ is the request complexity such that $log_2(c)$ is the request difficulty in bits.
+- $n$ is the number of active requests.
+- $g$ is the growth rate.
```
bobbinth, line 57: Should we mention that $g$ is a floating-point value?

docs/src/getting-started/cli.md
```diff
@@ -51,7 +51,8 @@ miden-faucet start \
 | `--pow-baseline` | Base PoW difficulty (0-32). It's the starting difficulty when no requests are pending | `12` | No |
 | `--pow-challenge-lifetime` | Challenge validity duration, i.e. how long challenges remain valid. This affects the rate limiting, since it works by rejecting new submissions while the previous submitted challenge is still valid | `30s` | No |
 | `--pow-cleanup-interval` | Cache cleanup interval, i.e. how often expired challenges are removed | `2s` | No |
-| `--pow-growth-rate` | Difficulty growth rate, i.e. how quickly difficulty increases with load. When set to 1, the difficulty will roughly double when the number of requests doubles. | `1` | No |
+| `--pow-growth-rate` | Difficulty growth rate, i.e. how quickly difficulty increases with load. | `0.1` | No |
+| `--pow-base-difficulty-amount` | Token amount (in base units) at which the difficulty of the challenge starts to increase. | `100000000` | No |
```
bobbinth, line 55: This should be just `base-amount`, right?

bin/faucet/src/main.rs
```diff
@@ -113,22 +113,34 @@ pub enum Command {
...
         pow_cleanup_interval: Duration,
 
         /// The baseline for the `PoW` challenges. This sets the `PoW` difficulty (in bits) that a
         /// a challenge will have when there are no requests against the faucet. It must be between
-        /// 0 and 32.
+        /// 0 and 64.
         #[arg(value_parser = clap::value_parser!(u8).range(0..=32))]
-        #[arg(long = "pow-baseline", value_name = "U8", env = ENV_POW_BASELINE, default_value = "12")]
+        #[arg(long = "pow-baseline", value_name = "U8", env = ENV_POW_BASELINE, default_value = "16")]
         pow_baseline: u8,
```
bobbinth, line 134: Is `64` on line 131 correct? I think the valid range is still `0..32`.

bin/faucet/src/api/get_tokens.rs
```diff
@@ -238,8 +238,15 @@ impl RawMintRequest {
         // Validate Challenge and nonce
         let challenge_str = self.challenge.ok_or(MintRequestError::MissingPowParameters)?;
         let nonce = self.nonce.ok_or(MintRequestError::MissingPowParameters)?;
-
-        server.submit_challenge(&challenge_str, nonce, account_id, api_key.unwrap_or_default())?;
+        let request_complexity = (asset_amount.base_units() / server.metadata.base_amount) + 1;
```
bobbinth, line 241: I think there are a couple of places where we compute request complexity based on the amount. It may be good to have a small helper function for this so that in case the formula changes, we have just one place to update.

bobbinth (approved):
Looks good! Thank you! I left some small comments inline. Once these are addressed, this should be good to merge.

PR #82, opened 2025-09-09
fix: display minted token amount

TomasArrachea:
Currently there is a bug on the frontend that causes the minted amount to show as 0:
image

The problem is on how we index the selected option in the frontend.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 0.11.3 (2025-09-08)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.11.3 (2025-09-08)
```
bobbinth, line 3: nit: I think this should be `2025-09-90`.

PR #86, opened 2025-09-16
feat: instrument faucet run

TomasArrachea:
addresses https://github.com/0xMiden/miden-faucet/issues/85


igamigo (approved):
On a quick look, the error seems to be returned from this function when running `new_transaction`, so unless I'm missing something, this should be enough for Honeycomb to pick up the errors

TomasArrachea:
Should this PR go to `main`?

bobbinth (approved):
Looks good! Thank you! I'll merge it now and once we update the batch size will make the next release.

@TomasArrachea - could you make a quick PR to update the batch size and increment the version number?

PR #87, opened 2025-09-16
feat: set batch size to 8

TomasArrachea:
context: https://github.com/0xMiden/miden-faucet/issues/85

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.11.4 (2025-09-16)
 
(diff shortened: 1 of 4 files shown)


bobbinth (approved):
Looks good! Thank you!

mmagician (approved):
LGTM, I'd just reword the changelog entry a bit

mmagician:
For a separate PR, I wonder if setting the batch size would make sense via CLI arguments rather than hardcoding it?

bobbinth:
> For a separate PR, I wonder if setting the batch size would make sense via CLI arguments rather than hardcoding it?

Yes, let's create an issue for this.

PR #89, opened 2025-09-16
feat: set batching size as cli argument

### LINKED ISSUE #88: Set the batch size via CLI arguments
For a separate PR, I wonder if setting the batch size would make sense via CLI arguments rather than hardcoding it?

_Originally posted by @mmagician in https://github.com/0xMiden/miden-faucet/issues/87#issuecomment-3300170204_

TomasArrachea:
closes #88


TomasArrachea:
Pointing this to the https://github.com/0xMiden/miden-faucet/pull/87 PR in case we want to introduce this in the release, it might be useful.

crates/faucet/src/lib.rs
```diff
@@ -187,10 +186,12 @@ impl Faucet {
     pub async fn run(
         mut self,
         mut requests: Receiver<(MintRequest, MintResponseSender)>,
+        batch_size: usize,
     ) -> anyhow::Result<()> {
         let mut buffer = Vec::new();
 
-        while requests.recv_many(&mut buffer, BATCH_SIZE).await > 0 {
+        // TODO: batching 10 or more requests fails, see https://github.com/0xMiden/miden-faucet/issues/85
+        while requests.recv_many(&mut buffer, batch_size).await > 0 {
```
bobbinth, line 194: The issue should now be fixed - though, would be good to confirm this. If all looks good, we should remove the `TODO` and probably increase the default to something like 32 (or maybe even more) notes per tx.

docs/src/getting-started/cli.md
```diff
@@ -60,6 +60,7 @@ miden-faucet start \
 | `--remote-tx-prover-url` | Remote transaction prover | - | No |
 | `--api-keys` | Comma-separated API keys | - | No |
 | `--enable-otel` | Enable OpenTelemetry | `false` | No |
+| `--batch-size` | Maximum amount of requests to process in each batch. It sets how many p2id notes to create in each transaction | `8` | No |
```
bobbinth, line 63: nit: I'd write this as "Maximum number of P2ID notes to create per transaction."

bobbinth (approved):
Looks good! Thank you! I left a couple of small comments inline.

PR #90, opened 2025-09-17
chore: bump crate versions

mmagician:
draft PR until base & client are released.
Then: `cargo update -p miden-client`

I didn't modify the batch size here, might be worth bumping that up as well

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## 0.11.5 (2025-09-18)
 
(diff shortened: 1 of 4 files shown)


igamigo:
Because [this](https://github.com/0xMiden/miden-faucet/pull/89) is now targeting `next`, the batch size change should be rolled back cc @TomasArrachea

bobbinth (approved):
Looks good! Thank you! I've pushed a commit to refresh Cargo.lock and also increased batch size to 64.

PR #95, opened 2025-09-19
feat: add metadata polling

TomasArrachea:
closes https://github.com/0xMiden/miden-faucet/issues/45

### DIFF
--- a/bin/faucet/frontend/index.js
+++ b/bin/faucet/frontend/index.js
@@ -69,10 +69,10 @@ class MidenFaucet {
     startMetadataPolling() {
         this.fetchMetadata();
 
-        // Poll every 1 second
+        // Poll every 2 seconds
         this.metadataInterval = setInterval(() => {
             this.fetchMetadata();
-        }, 1000);
+        }, 2000);
     }
 
     async fetchMetadata() {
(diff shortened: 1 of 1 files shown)


bin/faucet/frontend/index.js
```diff
@@ -65,17 +66,31 @@ class MidenFaucet {
         }
     }
 
+    startMetadataPolling() {
+        this.fetchMetadata();
+
+        // Poll every 1 second
+        this.metadataInterval = setInterval(() => {
+            this.fetchMetadata();
+        }, 1000);
```
bobbinth, line 75: I wonder if 1 second may be a bit too aggressive. I think we can probably do 2 - 3 seconds without degrading user experience too much.

TomasArrachea: Ok! Increased it to 2 seconds.

bobbinth:
> I only looked into this briefly but I think that for this simple case where we just return few metadata values and latency is not an issue, is not really worth it to use long-lived connections.

Agreed - though, I'd still create an issue because it may be good to work though the general approach of how such updates could be communicated to the front-end.

TomasArrachea:
> Agreed - though, I'd still create an issue because it may be good to work though the general approach of how such updates could be communicated to the front-end

Created https://github.com/0xMiden/miden-faucet/issues/97. 

It might be more efficient for the server to use SSE upon heavy load, but we should double check this. Anyway, we are already using SSE to send the updates during the minting process.

PR #96, opened 2025-09-19
feat: get issuance from the store

TomasArrachea:
Addresses https://github.com/0xMiden/miden-faucet/issues/44.


crates/faucet/src/lib.rs
```diff
@@ -298,11 +305,26 @@ impl Faucet {
...
+            .client
+            .get_account(self.id.account_id)
+            .await?
+            .context("faucet account not found")?
+            .account()
+            .get_token_issuance()
+            .context("faucet account is not a valid fungible faucet account")?;
+
+        Ok(AssetAmount::new(issuance.as_int())?)
     }
```
bobbinth, line 324: This works, but also, reading the full account from the database just to read a single storage slot seems kind of expensive. An alternative could be to introduce a method for reading specific storage slots - something like:
```rust
impl Client {

    pub fn get_storage_item(
        &self,
        account_id: AccountId,
        slot_index: u8,
    ) -> Result<Word, SomeError> {
        ...
    }
}
```

But a few open questions:
- Using `slot_index` here is somewhat fragile because it requires making assumptions about the storage layout. This problem should go away after we move to named storage slots.
- How to handle different storage slot types? For example, would this return an error if the underlying storage slot is a storage map?
- Is this the best interface? Maybe we should have something like `StorageReader` which can be used to simplify storage reading of an account. For example:

```rust
impl Client {

    pub fn read_storage(
        &self,
        account_id: AccountId,
    ) -> Result<StorageReader, SomeError> {
        ...
    }
}
```
And then we could do something like:
```rust
let issuance = self
    .client
    .read_storage(self.id.account_id)
    .context("faucet account not found")?
    .get_item("faucet::issuance")
    .context("issuance storage item not found")?;
    // TODO: somehow extract issuance from the storage item
```

cc @igamigo and @mmagician in case they have any thoughts on this.

mmagician: > Maybe we should have something like StorageReader which can be used to simplify storage reading of an account. For example:

I think this approach would work well once we have named storage slots.
For now I'd actually keep the current implementation though (since we expect to get named storage slots soon ™️ and this would need to be revisited)

crates/faucet/src/lib.rs
```diff
@@ -471,13 +483,27 @@ impl Faucet {
...
+
+    /// Get the current issuance of the faucet by querying the client's store.
+    pub async fn get_issuance(&self) -> anyhow::Result<AssetAmount> {
+        let faucet_sysdata = self
+            .client
+            .new_storage_reader(self.id.account_id)
+            .get_item(AccountStorage::faucet_sysdata_slot().clone())
+            .await
+            .context("failed to get faucet sysdata slot")?;
+        let issuance_felt = faucet_sysdata[Account::ISSUANCE_ELEMENT_INDEX];
```
TomasArrachea: This will be improved once [this issue](https://github.com/0xMiden/miden-base/issues/2344) is tackled in miden-base

PR #98, opened 2025-09-23
feat: add batching test

TomasArrachea:
Addresses https://github.com/0xMiden/miden-faucet/issues/92

This PR introduces a test to the `miden-faucet-lib` crate that validates that the faucet is able to batch 32 requests and generate a transaction that creates 32 p2id notes.


igamigo (approved):
LGTM! Is there an easy way of asserting the amount of notes created by the transaction? Maybe by looking at the mock RPC component? If so, that'd be nice as well.

igamigo (approved):

PR #99, opened 2025-09-23
feat: remove string encoding from challenge

TomasArrachea:
Closes https://github.com/0xMiden/miden-faucet/issues/83

### DIFF
--- a/docs/src/api-usage.md
+++ b/docs/src/api-usage.md
@@ -26,16 +26,17 @@ const target: bigint = BigInt(powJson.target);
 ```typescript
 // Dependencies: npm i @noble/hashes
 import { sha_256 } from '@noble/hashes/sha2.js';
-import { utf8ToBytes } from '@noble/hashes/utils.js';
+import { hexToBytes } from '@noble/hashes/utils.js';
 
 let nonce = 0;
+const challengeBytes = hexToBytes(challenge);
 while (true) {
     nonce = Math.floor(Math.random() * Number.MAX_SAFE_INTEGER);
 
     try {
         // Compute hash using SHA-256 with the challenge and nonce
         let hash = sha_256.create();
-        hash.update(utf8ToBytes(challenge)); // Use the hex-encoded challenge string directly
+        hash.update(challengeBytes);
 
         // Convert nonce to 8-byte big-endian format to match backend
         const nonceBytes = new ArrayBuffer(8);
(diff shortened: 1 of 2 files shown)


crates/pow/Cargo.toml
```diff
@@ -15,10 +15,10 @@ version.workspace      = true
 workspace = true
 
 [dependencies]
-serde     = { workspace = true }
-sha2      = { workspace = true }
-thiserror = { workspace = true }
-tokio     = { features = ["fs"], workspace = true }
+miden-client = { workspace = true }
```
bobbinth, line 18: Could we avoid introducing `miden-client` dependency here? I think `Challenge` struct is simple enough where we can probably serialize it "by hand" and it may actually end up being less code. For example, we could have something like:

```rust
impl Challenge {
    pub fn to_bytes(&self) -> [u8; SERIALIZED_SIZE] {
        ...
    }
}

impl From<Challenge> for [u8;  SERIALIZED_SIZE] {
    ...
}

impl TryFrom<[u8; SERIALIZED_SIZE]> for Challenge {
    ...
}

impl TryFrom<&[u8]> for Challenge {
    ...
}
```

crates/pow/Cargo.toml
```diff
@@ -15,10 +15,10 @@ version.workspace      = true
...
+sha2         = { workspace = true }
+thiserror    = { workspace = true }
+tokio        = { features = ["time"], workspace = true }
```
bobbinth, line 21: Not for this PR, but I wonder if it may make sense to make `tokio` an optional dependency - i.e., maybe the user of the rate-limiter could control how the cleanup is invoked. Let's create an issue for this.

bobbinth (commented):
Looks good! Thank you! I left a couple more small comments inline.

Also, do we need to update any docs because of this change? e.g., docs for how we decode the challenge in JS.

TomasArrachea:
> Last question: do we need to update the JS/TS docs and examples to make sure they reflect how challenges are serialized, or are they good as is?

This was actually missing, added it. Thanks!

bin/faucet/src/api/get_pow.rs
```diff
@@ -16,13 +17,25 @@ use crate::error_report::ErrorReport;
...
     let mut requestor = [0u8; 32];
     requestor[..AccountId::SERIALIZED_SIZE].copy_from_slice(&account_id_bytes);
     let challenge = rate_limiter.build_challenge(requestor, request.api_key);
-    Ok(Json(challenge))
+
+    Ok(Json(GetPowResponse {
+        challenge: challenge.to_bytes().to_hex(),
+        target: challenge.target(),
+        timestamp: challenge.timestamp(),
+    }))
```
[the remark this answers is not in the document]
TomasArrachea: Correct.

PR #103, opened 2025-09-26
feat: show PoW difficulty on the frontend

### LINKED ISSUE #78: Show PoW difficulty in the UI
After https://github.com/0xMiden/miden-faucet/issues/42, the requested token amount will affect the PoW challenge difficulty. We should show in the UI the estimated difficulty/time that each option will require, so the user knows beforehand.

To implement this, we could extend the faucet metadata to contain the current difficulty -that depends on the faucet load- and then the frontend computes how much difficulty each option adds. Also, the metadata should contain the `pow_baseline` so that the faucet can use the difficulty to compute the estimated time.

TomasArrachea:
Closes #78 

This PR updates the frontend to show a hint on the estimated time depending on which amount of tokens is selected. To implement it, the metadata was extended to include the load difficulty. Then the frontend can compute the amount of difficulty bits and uses it to show an estimated time. The estimations are:
- 17 difficulty bits -> "<5s"
- 18 difficulty bits -> "5s-15s"
- 19 difficulty bits -> "15-30s"
- 20 difficulty bits -> "30s-1m"
- 21 difficulty bits -> "1m-5m"
- 22+ difficulty bits -> "5m+"

Preview: 

https://github.com/user-attachments/assets/e47c19db-67be-4f72-99af-7794733bc0a0

### DIFF
--- a/bin/faucet/frontend/index.js
+++ b/bin/faucet/frontend/index.js
@@ -23,14 +23,28 @@ class MidenFaucet {
             console.error("Web Crypto API not available");
             this.showError('Web Crypto API not available. Please use a modern browser.');
         }
-
-        this.startMetadataPolling();
         this.privateButton.addEventListener('click', () => this.handleSendTokens(true));
         this.publicButton.addEventListener('click', () => this.handleSendTokens(false));
         this.walletConnectButton.addEventListener('click', () => this.handleWalletConnect());
         this.tokenSelect.addEventListener('change', () => this.updateTokenHint());
 
         this.walletAdapter = new MidenWalletAdapter({ appName: 'Miden Faucet' });
+
+        this.init();
+    }
+
+    async init() {
+        this.backendUrl = await this.getBackendUrl();
+        this.startMetadataPolling();
+    }
+
+    async getBackendUrl() {
+        const response = await fetch('/config.json');
+        if (!response.ok) {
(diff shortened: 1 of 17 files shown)


bin/faucet/frontend/index.js
```diff
@@ -354,6 +359,33 @@ class MidenFaucet {
...
+        } else if (difficultyBits <= 19) {
+            estimatedTime = `15-30s`;
+        } else if (difficultyBits <= 20) {
+            estimatedTime = `30s-1m`;
+        } else if (difficultyBits <= 21) {
+            estimatedTime = `1-5m`;
+        } else {
+            estimatedTime = `5m+`;
+        }
+        this.tokenAmountHint.textContent = `Larger claims involve longer processing times. Estimated: ${estimatedTime}`;
```
TomasArrachea, line 387: I'm not sure about this phrasing

bobbinth: Maybe "Larger amounts take more time to mint. Estimated: ..."?

mmagician (approved):
Where do these estimates come from?
```
17 difficulty bits -> "<5s"
18 difficulty bits -> "5s-15s"
19 difficulty bits -> "15-30s"
20 difficulty bits -> "30s-1m"
21 difficulty bits -> "1m-5m"
22+ difficulty bits -> "5m+"
```

Otherwise LGTM

PR #110, opened 2025-10-07
feat: redesign frontend

TomasArrachea:


TomasArrachea:
>One item we might want to think about is the gray progress bar: if things progress quickly, it's pretty clear, but when the processing is slow it actually looks like a redundant horizontal scroll bar:

Currently this bar only loads halfway once the PoW challenge is solved and then to the end once the minting is complete, it doesn't progress slowly. I can think 2 options:
1. Make the bar progress linearly with the estimations made in https://github.com/0xMiden/miden-faucet/pull/103 (and maybe add some random jitter for display purposes) .

2. Replace the bar it with steps numbers, something like: `1/3 PREPARING THE REQUEST` -> `2/3 MINTING TOKENS` -> `3/3 TOKENS MINTED!`

In any case, we can move forward with this PR and leave that as a follow-up.

bobbinth (approved):
Looks good! Thank you! This is a pretty superficial review from me - but I think we can merge, deploy, and see how it looks on devnet.

PR #116, opened 2025-10-09
feat: improve error reports

### LINKED ISSUE #115: Add detail to PoW error trace
Currently the `get_tokens` endpoint logs the `GetTokenError` errors encountered. When the challenge is invalid, the error showed is: `invalid request: PoW error`. We could improve this a bit so the inner error is also logged to see if the error is because of rate limiting or due to invalid challenges.

TomasArrachea:
closes #115


mmagician (approved):
LGTM ✅

PR #118, opened 2025-10-13
feat: remove local prover fallback

TomasArrachea:
closes https://github.com/0xMiden/miden-faucet/issues/112


TomasArrachea:
Ok, given that the testnet machine is not specially powerful and proving should not take too long on other machines, then we can keep things as is and close this for now.

PR #119, opened 2025-10-14
feat: separate backend and frontend servers

TomasArrachea:
closes https://github.com/0xMiden/miden-faucet/issues/17


bin/faucet/src/backend/get_metadata.rs
```diff
@@ -12,6 +12,7 @@ use url::Url;
 use crate::COMPONENT;
 
 /// Describes the faucet metadata needed to show on the frontend.
+#[derive(Clone)]
 pub struct Metadata {
```
igamigo: minor nit: Not from this PR but feels like `Metadata` is too general. Not sure what a good alternative would be though, since `FaucetMetadata` is a bit misleading considering it also has the explorer URL

igamigo (approved):
LGTM, just one very minor comment

PR #126, opened 2025-10-16
feat: sync state before creating the minting transaction

TomasArrachea:
As discussed offlined with @igamigo:

The testnet faucet started failing due to a pair of transactions not being submitted correctly to the node (still unknown why). This caused an out of sync state, where all subsequent transactions failed to submit to the node due to having a different account commitment.

This PR changes to sync the state before creating the mint transaction, so we ensure that the state is synced.


TomasArrachea:
>I don't see any downsides to this - are there any?

I think "cold starts" will be a bit slower, since now the first batch of requests will have to wait for the sync. Before, the batches were handled and only then the sync took place. But since the queue usually contains requests that are essentially waiting for the previous sync, response times should stay the same. Anyway, the sync was improved and now its taking a lot less time.

PR #127, opened 2025-10-16
feat: download only commited notes

TomasArrachea:
closes https://github.com/0xMiden/miden-faucet/issues/22.
Addresses part of https://github.com/0xMiden/miden-faucet/issues/5

This PR uses the `RpcClient` from `@demox-labs/miden-sdk` to poll notes from the node given their ids. This is used to confirm that the minting transactions have been committed to the node. 

For this, a new step was added in the loading modal. This can potentially make the minting process slower, but the user would have to wait for it anyway.

Preview:

https://github.com/user-attachments/assets/692ed963-9de6-4365-97e3-0ed70add3a99


bin/faucet/src/frontend.rs
```diff
@@ -47,6 +48,17 @@ pub async fn get_index_html() -> Html<&'static str> {
...
+    (
+        [(header::CONTENT_TYPE, header::HeaderValue::from_static("application/wasm"))],
+        include_bytes!(concat!(
+            env!("CARGO_MANIFEST_DIR"),
+            "/frontend/node_modules/@demox-labs/miden-sdk/dist/assets/miden_client_web.wasm"
+        ),),
+    )
+        .into_response()
+}
+
```
TomasArrachea: This is needed to make the wasm available in the bundle. Not very pretty but I couldn't make it work using bundler settings or plugins.

bin/faucet/frontend/index.js
```diff
@@ -398,6 +384,50 @@ class MidenFaucet {
...
+        };
+    }
+
+    async setupDownloadButton(noteId) {
+        const downloadButton = document.getElementById('download-button');
+        downloadButton.onclick = async () => {
+            await this.downloadNote(noteId);
+            this.showCloseButton();
+        };
+    }
```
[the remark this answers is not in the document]
TomasArrachea: Ok, I agree we could refactor this js to be nicer. Let's do it in a separate PR.

bin/faucet/src/frontend.rs
```diff
@@ -29,6 +30,8 @@ pub async fn serve_frontend(url: Url, api_url: Url) -> anyhow::Result<()> {
         .route("/wallet-icon.png", get(get_wallet_icon))
         .route("/favicon.ico", get(get_favicon))
         .route("/config.json", get(config_json))
+        // TODO: this is a hacky way of making the web-client work. Ideally the wasm file should be included in the bundle.
```
juan518munoz: Should we open an issue to track this?

PR #129, opened 2025-10-20
feat: update pow estimations

TomasArrachea:
This PR slightly updates the pow estimations added In https://github.com/0xMiden/miden-faucet/pull/103.

To make the estimations I had ran a small benchmark in my browser, but the results had high variance. Here I changed the benchmark a bit to use a deterministic set of iterations per challenge and got more stable results using an expected value of iterations (`E(X) = 2 ^ leading_zeros`). 

Still, these values are just for approximations.

### DIFF
--- a/bin/faucet/frontend/index.js
+++ b/bin/faucet/frontend/index.js
@@ -383,18 +383,18 @@ class MidenFaucet {
         const difficultyBits = Math.log2(difficulty);
 
         let estimatedTime;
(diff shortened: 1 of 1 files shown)


TomasArrachea:
Thanks for running it, here is the updated js code I ran in this PR to reduce the variance: https://gist.github.com/TomasArrachea/0b878a40e55e56e68623bcf2ddcf1c9d

PR #132, opened 2025-10-23
feat: sync state before creating the minting transaction

TomasArrachea:
Cherry-picked commit https://github.com/0xMiden/miden-faucet/commit/ebed37fe656c6940ce103f7c00e78131050acb25 from next into main.

This prevents the faucet to fail when it desyncs it's state with the node, as explained in the original [PR](https://github.com/0xMiden/miden-faucet/pull/126).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1779,7 +1779,7 @@ dependencies = [
 
 [[package]]
 name = "miden-faucet"
-version = "0.11.7"
+version = "0.11.8"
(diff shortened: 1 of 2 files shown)


igamigo:
Yeah, ideally we should release a new version. This has happened a couple of times in the last few weeks (likely related to the node crashes BTW) so it'd be nice to fix it in the current testnet.

PR #141, opened 2025-10-31
feat: use sync component

TomasArrachea:
This PR uses the client's `SyncState` component to make the syncing more custom for the faucet use case.

In particular:
- no note tags are sent on the sync request. This avoids matching with unnecessary blocks due to the prefixes collisions.
- only sync expected output notes instead of unconsumed. This means we don't sync notes that already were marked as commited. Otherwise they would be synced to check for consumption, but we don't care.

### DIFF
--- a/crates/faucet/src/lib.rs
+++ b/crates/faucet/src/lib.rs
@@ -209,7 +209,7 @@ impl Faucet {
             .sync_state(
                 current_partial_mmr,
                 accounts,
-                note_tags.clone(),
+                note_tags,
                 input_notes,
                 expected_output_notes,
                 uncommitted_transactions,
--- a/crates/faucet/src/note_screener.rs
+++ b/crates/faucet/src/note_screener.rs
@@ -1,5 +1,5 @@
-//! Implements a custom note screener that only marks notes as relevant if they are tagged. It
-//! discards all other notes.
+//! Implements a custom note screener that only marks notes as relevant if they are tracked output
+//! notes. It discards all other notes.
 use std::sync::Arc;
 
 use miden_client::rpc::domain::note::CommittedNote;


crates/faucet/src/lib.rs
```diff
@@ -180,13 +180,68 @@ impl Faucet {
...
+        info!("Applying changes to the store.");
+
+        // Apply received and computed updates to the store
+        store
+            .apply_state_sync(state_sync_update)
+            .await
+            .map_err(ClientError::StoreError)?;
+
+        // Remove irrelevant block headers
+        store.prune_irrelevant_blocks().await?;
```
igamigo: I think this is not needed (currently done as part of `apply_state_sync()`

crates/faucet/src/lib.rs
```diff
@@ -180,13 +180,68 @@ impl Faucet {
...
+        // Get current state of the client
+        let accounts = store
+            .get_account_header(account_id)
+            .await?
+            .map(|(header, _)| vec![header])
+            .unwrap_or_default();
+        let note_tags: BTreeSet<NoteTag> = store.get_unique_note_tags().await?;
+
+        let unspent_input_notes = vec![];
+        let unspent_output_notes = vec![];
```
igamigo: Here, I think we do want to update notes that were not committed:


```suggestion
        let expected_output_notes = store.get_output_notes(OutputNoteFilter::Expected).await?;
```

crates/faucet/Cargo.toml
```diff
@@ -16,6 +16,7 @@ workspace = true
 
 [dependencies]
 # Miden dependencies.
+async-trait               = { version = "0.1" }
```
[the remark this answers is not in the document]
TomasArrachea: It seems to be needed. I'm not sure why but removing the marker gives a compilation error.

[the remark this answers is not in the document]
TomasArrachea: There was no simple way of redefining the trait to not need the `async_trait(?Send)` here. Added the reexport of `async-trait` in the client

crates/faucet/src/note_screener.rs
```diff
@@ -0,0 +1,45 @@
+// Implement a custom note screener that only marks notes as relevant if they are tagged. It
```
igamigo: nit: Would this be better placed as a module doc comment?

igamigo (approved):
LGTM! Would be great to get a review from @mmagician as well

crates/faucet/src/lib.rs
```diff
@@ -170,13 +174,59 @@ impl Faucet {
...
+    ) -> anyhow::Result<SyncSummary> {
+        // Get current state of the client
+        let accounts = client
+            .get_account_header_by_id(account_id)
+            .await?
+            .map(|(header, _)| vec![header])
+            .unwrap_or_default();
+        let note_tags = BTreeSet::new();
+        let input_notes = vec![];
+        let expected_output_notes = client.get_output_notes(NoteFilter::Expected).await?;
```
mmagician: Oh and `Expected` vs. `Unspent`

PR #147, opened 2025-11-10
feat: add cli tests

TomasArrachea:
closes https://github.com/0xMiden/miden-faucet/issues/139

Adds test to the CLI after the refactor introduced in https://github.com/0xMiden/miden-faucet/pull/84

### DIFF
--- a/bin/faucet/src/main.rs
+++ b/bin/faucet/src/main.rs
@@ -594,7 +594,7 @@ mod tests {
         let result = Box::pin(run_faucet_command(Cli::parse_from([
             "miden-faucet",
             "start",
-            "--api-url",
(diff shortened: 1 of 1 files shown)


bin/faucet/Cargo.toml
```diff
@@ -58,6 +58,7 @@ reqwest          = { default-features = false, features = ["json"], version = "0
 serde_json       = { workspace = true }
 tokio            = { features = ["macros", "process"], workspace = true }
 tonic-web        = { version = "0.14" }
+uuid             = { features = ["v4"], version = "1.10" }
```
mmagician: as this import is now used by two crates, let's make it a `workspace` dependency

PR #151, opened 2025-11-11
fix: cargo packaging (track `bundle.js` & `miden_client_web.wasm` in git)

mmagician:
- adds `bin/faucet/frontend/bundle.js` to git
- adds `bin/faucet/frontend/miden_client_web.wasm` to git
    - this file is copied from `node_modules/@demox-labs/miden-sdk/dist/assets/miden_client_web.wasm` when running `make build`
    - changes `bin/faucet/src/frontend.rs` to reference `miden_client_web.wasm` from `bin/faucet/frontend/` instead of `node_modules/@demox-labs/miden-sdk/dist/assets/`
- Makefile changes: `build` and `install-faucet` commands now both run `cd bin/faucet/frontend && npm install && npm run build && cp node_modules/@demox-labs/miden-sdk/dist/assets/miden_client_web.wasm .`
    - Question: is it enough to only have `make build` run it?
- CI workflow to test for drift in the committed `bundle.js` & `miden_client_web.wasm`
- tracks `package-lock.json` files in git, needed for `npm ci` (clean install)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,9 +17,8 @@ target
 keystore/
 faucet_client_store.sqlite3
 
-# Node files used for examples
+# Node files
 node_modules/
-package-lock.json
 
 # Ignore mdbook build files
 book/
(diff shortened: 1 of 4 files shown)


igamigo:
Why do we need this? From the PR description it's not immediately clear. Is the problem that the `build.rs` would also write under `./src`?

PR #153, opened 2025-11-11
feat: add `api-public-url` parameter

TomasArrachea:
Closes https://github.com/0xMiden/miden-faucet/issues/152.

This PR adds a new `--api-public-url` parameter that sets the public API url to bake into the frontend config.json.


docs/src/getting-started/cli.md
```diff
@@ -107,7 +108,8 @@ export MIDEN_FAUCET_MAX_SUPPLY=
 
 # Faucet Service Configuration
 export MIDEN_FAUCET_FRONTEND_URL=http://localhost:8080
-export MIDEN_FAUCET_API_URL=http://localhost:8000
+export MIDEN_FAUCET_API_BIND_URL=http://localhost:8000
```
igamigo: Would be nice to tackle https://github.com/0xMiden/miden-faucet/issues/152#issuecomment-3518547757, unless we think there's some reason not to right now

PR #154, opened 2025-11-11
feat: build frontend artifacts in OUT_DIR

TomasArrachea:
This aims to fix the release failure that appeared in https://github.com/0xMiden/miden-faucet/actions/runs/19250439397.

This PR updates the build script to copy all the frontend files into OUT_DIR, and then runs `npm install` and `npm build` to generate the needed artifacts. So, the build script does not generate changes on the source directory.

Related: https://github.com/0xMiden/miden-faucet/pull/151


igamigo (approved):
Thanks! If `cargo package` and `cargo publish --dry-run` work with this, I believe we should go for this rather than #151

EDIT: Tried it and it works

bobbinth (approved):
Looks good! Thank you! Assuming `cargo publish --dry-run` works with this, I think we should go with this approach.

PR #156, opened 2025-11-12
feat: improve CLI server config params

TomasArrachea:
Closes https://github.com/0xMiden/miden-faucet/issues/163

This PR:
- Replaced the `api-url` param for `api-bind-port` and always binds on `http://0.0.0.0:<api-bind-port>`. The param is now optional, using 8000 as default.
- Replaces the `frontend-url` param for `frontend-bind-port` and always binds on `http://0.0.0.0:<frontend-bind-port>`. The param is now optional, using 8080 as default.
- Adds `no-frontend` param to optionally disable the frontend server. Previously this was done when no `frontend-url` was set, but now that param takes a default value.
- Updates the README and docs.

Now, running the faucet is as simple as:
```
miden-faucet init \
  --token-symbol MIDEN \
  --decimals 6 \
  --max-supply 100000000000000000 \
  --network devnet
```

```
miden-faucet start --network devnet
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 - [BREAKING] Replaced the `frontend-url` param for `frontend-bind-port` ([#156](https://github.com/0xMiden/miden-faucet/pull/156)).
 - [BREAKING] Added `no-frontend` param to optionally disable the frontend server ([#156](https://github.com/0xMiden/miden-faucet/pull/156)).
 
-## 0.12.4 (2025-11-27)
 ## 0.12.4 (2025-12-04)
 
 - Added version to the metadata endpoint ([#169](https://github.com/0xMiden/miden-faucet/pull/169)).


docs/src/getting-started/cli.md
```diff
@@ -64,8 +64,8 @@ miden-faucet start \
 
 | Option | Description | Default | Required |
 |--------|-------------|---------|----------|
-| `--api-bind-url` | URL to serve the faucet API | - | Yes |
-| `--api-public-url` | Public URL to access the faucet API. If not set, the bind url will be used. | - | No |
+| `--api-port` | Port to bind the API server | - | Yes |
+| `--api-public-url` | Public URL to access the faucet API | - | No |
```
bobbinth: nit (feel free to ignore): I wonder if `api-bind-port` would be a bit more self-explanatory.

SantiagoPittella (approved):

bin/faucet/src/main.rs
```diff
@@ -130,17 +131,21 @@ pub enum Command {
...
-        /// URL to bind the frontend server. If not set, the frontend will not be served.
-        #[arg(long = "frontend-url", value_name = "URL", env = ENV_FRONTEND_URL)]
-        frontend_url: Option<Url>,
+        /// Port to bind the frontend server.
+        #[arg(long = "frontend-bind-port", value_name = "PORT", env = ENV_FRONTEND_BIND_PORT, default_value = "8080")]
+        frontend_bind_port: u16,
+
+        /// Optionally disable the frontend server.
+        #[arg(long = "no-frontend", value_name = "BOOL", default_value_t = false, env = ENV_NO_FRONTEND)]
+        no_frontend: bool,
```
bobbinth: nit (and not for this PR): I wonder if "website" would be a slightly better name than "frontend". Let's create an issue for this.

bin/faucet/src/main.rs
```diff
@@ -130,17 +131,21 @@ pub enum Command {
         #[clap(flatten)]
         config: ClientConfig,
 
-        /// URL to bind the API server.
-        #[arg(long = "api-bind-url", value_name = "URL", env = ENV_API_BIND_URL)]
-        api_bind_url: Url,
+        /// Port to bind the API server.
+        #[arg(long = "api-bind-port", value_name = "PORT", env = ENV_API_BIND_PORT, default_value = "8000")]
+        api_bind_port: u16,
```
bobbinth: nit: should we add to the comment that the server will be started on `0.0.0.0`? Something like:

```
/// Port to bind the API server. The server will be started on `0.0.0.0:api-bind-port`.
```

bobbinth (approved):
Looks good! Thank you! I left a couple of small nits inline.

Also, I wonder if we should adapt the same methodology for specifying server binding addresses in the node (cc @Mirko-von-Leipzig).

bin/faucet/src/main.rs
TomasArrachea: created https://github.com/0xMiden/miden-faucet/issues/197

PR #167, opened 2025-11-15
feat: improve rate limiter

TomasArrachea:
This PR builds on top of https://github.com/0xMiden/miden-faucet/pull/165. Not sure if we want to point this to `main` or `next`, since is somewhat a bigger refactor.

While inspecting the rate limiter I found the following problem.

Basically the rate limiter indexes challenges by their timestamp. Users are rate-limited while their previous submitted challenge is still valid (its timestamp has not expired). But this allows a user to design a pipeline where they submit a challenge right before it is expired. Since there is no limit for requesting challenges, they could priodically submit challenges that are about to expire and thus not be rate limited at all.

To solve it, this PR changes the cache to track challenges by the timestamp of their submission rather than the timestamp of their creation.

### DIFF
--- a/crates/pow/src/challenge_cache.rs
+++ b/crates/pow/src/challenge_cache.rs
@@ -56,7 +56,8 @@ impl ChallengeCache {
     ) -> Result<(), ChallengeError> {
         let solver = (challenge.requestor, challenge.domain);
 
-        // Check if the solver is rate limited
+        // Check if the solver is rate limited. There could still be an expired challenge in the
+        // cache for this solver, so in that case we override it.
         let remaining_time = self.next_challenge_delay(&solver, current_time);
         if remaining_time != 0 {
             return Err(ChallengeError::RateLimited(remaining_time));
@@ -66,6 +67,10 @@ impl ChallengeCache {
 
         let prev_challenge = self.challenges_timestamps.insert(solver, current_time);
         if let Some(prev_timestamp) = prev_challenge {
+            assert!(
+                prev_timestamp + self.challenge_lifetime.as_secs() <= current_time,
+                "previous timestamp should be expired"
+            );
             // Since the previous timestamp for this solver is overridden, we can also just clean
             // up that challenge from the cache. The number of challenges for the domain stays
             // unchanged.
--- a/crates/pow/src/tests.rs
+++ b/crates/pow/src/tests.rs
@@ -275,7 +275,6 @@ async fn submit_challenge_while_previous_one_is_not_cleaned_up() {
         current_time,
         request_complexity,
     );
-    dbg!(&result);
     assert!(result.is_ok());
 
     // check that the first challenge is removed from the cache


crates/pow/src/challenge_cache.rs
```diff
@@ -42,24 +45,16 @@ impl ChallengeCache {
...
-        // Check if (timestamp, requestor, domain) is already in the cache.
-        let consumers = self.challenges.entry(challenge.timestamp).or_default();
-        if consumers.contains(&consumer) {
-            return false;
-        }
-        consumers.push(consumer);
+        self.challenges.entry(current_time).or_default().insert(consumer);
 
-        let prev_challenge = self.challenges_timestamps.insert(consumer, challenge.timestamp);
+        let prev_challenge = self.challenges_timestamps.insert(consumer, current_time);
```
TomasArrachea: This is the relevant change

crates/pow/src/lib.rs
```diff
@@ -227,7 +221,7 @@ mod tests {
         PoWRateLimiter::new(
             secret,
             PoWRateLimiterConfig {
-                challenge_lifetime: Duration::from_secs(30),
+                challenge_lifetime: Duration::from_secs(5),
```
juan518munoz: Any specific reason for this change?

TomasArrachea: This is used for testing, I added a test that needs to sleep for a `challenge_lifetime` so it's better for it to be short. Actually reduced it to 3 seconds.

bobbinth (commented):
Looks good! Thank you! I left a few small questions/comments inline.

Sorry it took such a long time to review.

crates/pow/src/lib.rs
```diff
@@ -176,15 +176,11 @@ impl PoWRateLimiter {
...
-        if !self
-            .challenges
+        // Insert the challenge into the cache
+        self.challenges
             .write()
             .expect("challenge cache lock should not be poisoned")
-            .insert_challenge(challenge)
-        {
-            return Err(ChallengeError::ChallengeAlreadyUsed);
-        }
```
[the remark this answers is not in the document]
TomasArrachea: I removed the error variant. I realized it does not make sense, since no challenge would be able to be submitted twice. While a challenge is valid, it would rate limit it's solver so it would always fail at the rate limiter check here: https://github.com/0xMiden/miden-faucet/blob/752d400aca8c2c7505d50a46c9593c0d129ae69c/crates/pow/src/lib.rs#L169-L177

When the submitted challenge gets cleaned up from the cache it's also expired, so it also cannot be submitted again.

crates/pow/src/challenge_cache.rs
```diff
@@ -42,30 +45,23 @@ impl ChallengeCache {
...
+    /// any.
     ///
-    /// Returns whether the challenge was newly inserted. That is:
-    /// * If the cache did not previously contain this challenge, `true` is returned.
-    /// * If the cache already contained this challenge, `false` is returned, and the cache is not
-    ///   modified.
-    pub fn insert_challenge(&mut self, challenge: &Challenge) -> bool {
+    /// Assumes that the consumer associated with the challenge is not rate limited. See
+    /// `next_challenge_delay`.
+    pub fn insert_challenge(&mut self, challenge: &Challenge, current_time: u64) {
```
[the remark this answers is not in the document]
TomasArrachea: Done in [7559284](https://github.com/0xMiden/miden-faucet/pull/167/commits/75592847bfa35ab06f184541d9527880cf428ff6)

PR #169, opened 2025-11-18
feat: expose version

TomasArrachea:
This PR updates the `/metadata` endpoint to expose the faucet version

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Install mdbook and plugins
         uses: taiki-e/install-action@v2
         with:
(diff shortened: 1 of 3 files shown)


bobbinth (approved):
Looks good! Thank you!

Technically, this is a breaking change, but I'm not sure if anyone aside from us uses the `/metadata` endpoint of the faucet - @Dominik1999, could you check with the relevant pioneers?

> Removed the linkcheck for now, maybe we can use this as an opportunity to migrate to docusaurus.

I'd probably keep the docs here in mdBook format - so, let's create an issue to figure out how to add back linkcheck later.

PR #174, opened 2025-11-26
feat: add web-client with cache and compression

### LINKED ISSUE #164: Re-add web-client
In https://github.com/0xMiden/miden-faucet/pull/158 we removed the web client because it was lagging the frontend.

We should re-add the removed functionality but being careful that the web-client is being correctly cached by the browser.

TomasArrachea:
closes #164 
closes https://github.com/0xMiden/miden-faucet/issues/140

This PR re-adds the web-client integration on the frontend.

Additionally:
- adds cache headers for the `bundle.js` and `miden_client_web.wasm` requests. This avoids having to request the resources repeteadly on each access to the website.
- adds compression to reduce the transfer size. This speeds the loading significantly for the first access. Now the WASM file request only transfers 3.2 kB (instead of 12 kB).

### DIFF
--- a/bin/faucet/src/main.rs
+++ b/bin/faucet/src/main.rs
@@ -637,6 +637,7 @@ mod tests {
         // Wait for the website to be fully loaded
         client
             .wait()
+            .at_most(Duration::from_secs(10))
             .for_element(fantoccini::Locator::Css("#token-amount option"))
             .await
             .unwrap();


bin/faucet/frontend/app.js
```diff
@@ -192,6 +200,54 @@ export class MidenFaucetApp {
...
+
+                    elapsedTime += currentInterval;
+
+                    if (elapsedTime <= 10000) {
+                        currentInterval = 500;
+                    } else if (elapsedTime <= 40000) {
+                        currentInterval = 1000;
+                    } else {
+                        currentInterval = 5000;
+                    }
```
juan518munoz: Not sure how this is handled in JS, but it may be more legible to have a `SECOND` and `MINUTE` declared as constants:
```js
const SECOND = 1000;
const MINUTE = 60 * SECOND;
```

So we can have this:
```suggestion
                    if (elapsedTime <= 10 * SECOND) {
                        currentInterval = 0.5 * SECOND;
                    } else if (elapsedTime <= 40 * SECOND) {
                        currentInterval = 1 * SECOND;
                    } else {
                        currentInterval = 5 * SECOND;
                    }
```

(Also applies to the rest of the code)

bin/faucet/frontend/app.js
```diff
@@ -192,6 +200,54 @@ export class MidenFaucetApp {
         }
     }
 
+    pollNote(noteId) {
+        return new Promise((resolve, reject) => {
+            // Poll every 500ms for the first 10 seconds, then every 1s for the next 30 seconds, then every 5s.
+            let currentInterval = 500;
+            let pollInterval;
```
juan518munoz: nit: prefer variable initialization
```suggestion
            let pollInterval = 0;
```

bobbinth:
Is there anything outstanding on this PR? Or does it just need reviews?

bin/faucet/src/main.rs
```diff
@@ -763,21 +768,6 @@ mod tests {
             });
         });
 
-        // Wait for faucet to be up
```
WiktorStarczewski: Quick question, why are we removing this?

WiktorStarczewski (commented):
Left comments. Sorry for the delay.

bin/faucet/frontend/app.js
```diff
@@ -192,6 +203,54 @@ export class MidenFaucetApp {
...
+                    clearInterval(pollInterval);
+                    pollInterval = setInterval(poll, currentInterval);
+                } catch (error) {
+                    console.error('Error polling for note:', error);
+                    clearInterval(pollInterval);
+                    clearTimeout(timeoutId);
+                    reject('Error fetching note confirmation.');
+                }
+            };
+            pollInterval = setInterval(poll, currentInterval);
```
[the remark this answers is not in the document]
TomasArrachea: Agreed it looks better, updated it in 78410a2

bin/faucet/frontend/app.js
```diff
@@ -192,6 +203,40 @@ export class MidenFaucetApp {
...
+                    console.error('Error polling for note:', err);
+                    return reject('Error fetching note confirmation.');
+                }
+
+                // choose next delay: 0.5s up to 10s, then 1s up to 40s, then 5s
+                const elapsed = Date.now() - start;
+                const nextDelay =
+                    elapsed <= 10 * SECOND ? 0.5 * SECOND :
+                        elapsed <= 40 * SECOND ? 1 * SECOND :
+                            5 * SECOND;
```
fkrause98: Maybe we can do this instead?
```suggestion
let nextDelay;
if (elapsed <= 10 * SECOND) {
    nextDelay = 0.5 * SECOND;
} else if (elapsed <= 40 * SECOND) {
    nextDelay = 1 * SECOND;
} else {
    nextDelay = 5 * SECOND;
}
```

PR #188, opened 2025-12-04
feat: make `tokio` optional for pow crate

TomasArrachea:
closes https://github.com/0xMiden/miden-faucet/issues/105


crates/pow/src/lib.rs
bobbinth: The docs for `run_cleanup()` function need to be updated (i.e., AFAICT, it doesn't run every minute but sleeps for the curation of `cleanup_interval`). And we should probably mention that this is a blocking function.

But also, maybe it is better to move `run_cleanup()` logic here and make `cleanup_expired_challenges()` public instead. This way, we make no assumptions about how the cleanup task is managed.

bobbinth (approved):
Looks good! Thank you!

PR #200, opened 2025-12-11
fix: publish dry-run

TomasArrachea:
In https://github.com/0xMiden/miden-faucet/pull/188 the cleanup task was put behind a `tokio` feature, but there a tokio::time::interval import left without the cfg marker that broke the dry-publish workflow.

### DIFF
--- a/crates/pow/src/lib.rs
+++ b/crates/pow/src/lib.rs
@@ -61,14 +61,7 @@ impl PoWRateLimiter {
 
             loop {
                 interval.tick().await;
-                let current_time = SystemTime::now()
-                    .duration_since(UNIX_EPOCH)
-                    .expect("current timestamp should be greater than unix epoch")
-                    .as_secs();
-                challenge_cache
-                    .write()
-                    .expect("challenge cache lock should not be poisoned")
-                    .drop_expired_challenges(current_time);
+                remove_expired_challenges(&challenge_cache);
             }
         });
 
@@ -93,14 +86,7 @@ impl PoWRateLimiter {
 
     /// Cleans up the challenge cache, removing all the expired challenges.
     pub fn cleanup_challenge_cache(&self) {
-        let current_time = SystemTime::now()
-            .duration_since(UNIX_EPOCH)
-            .expect("current timestamp should be greater than unix epoch")
-            .as_secs();
-        self.challenges
-            .write()
-            .expect("challenge cache lock should not be poisoned")
-            .drop_expired_challenges(current_time);
+        remove_expired_challenges(&self.challenges);
     }
 
     /// Generates a new challenge with a difficulty that will depend on the number of active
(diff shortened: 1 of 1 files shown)


crates/pow/src/lib.rs
```diff
@@ -83,26 +88,21 @@ impl PoWRateLimiter {
...
-                .as_secs();
-            challenges
-                .write()
-                .expect("challenge cache lock should not be poisoned")
-                .cleanup_expired_challenges(current_time);
-        }
+    /// Returns a clone of the challenge cache lock.
+    pub fn challenge_cache(&self) -> Arc<RwLock<ChallengeCache>> {
+        self.challenges.clone()
+    }
```
bobbinth: Why do we need to make the challenge cache visible externally?

TomasArrachea: Users that use the crate without the `tokio` feature would need to access the cache for clean up.

bobbinth: Ah - I see. And we cannot make the interface be `PoWRateLimiter::cleanup_challenge_cache()`?

TomasArrachea: Yes! made some changes in [662d552](https://github.com/0xMiden/miden-faucet/pull/200/commits/662d552c4c5a33e93291d17d15ac59d4bc59dc63), I think now it's cleaner.

crates/pow/src/lib.rs
```diff
@@ -49,22 +48,25 @@ pub struct PoWRateLimiterConfig {
...
+    /// challenges.
     #[cfg(feature = "tokio")]
     pub fn new_with_cleanup(secret: [u8; 32], config: PoWRateLimiterConfig) -> Self {
-        let challenge_cache = Arc::new(RwLock::new(ChallengeCache::new(config.challenge_lifetime)));
+        let cleanup_interval = config.cleanup_interval;
+        let rate_limiter = Self::new(secret, config);
 
         // Start the cleanup task
-        let cleanup_state = challenge_cache.clone();
+        let cleanup_rate_limiter = rate_limiter.clone();
```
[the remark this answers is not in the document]
TomasArrachea: > I'm not sure this'll work: we are running the cleanup task against one rate limiter but returning another, right? I don't think we can clone the rate-limiter here.

The `PoWRateLimiter` holds an `Arc<RwLock<ChallengeCache>>`, so we are cloning the Arc but referencing the same challenge cache instance.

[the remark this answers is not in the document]
TomasArrachea: Yes, that would also work. Made those changes in [d74e067](https://github.com/0xMiden/miden-faucet/pull/200/commits/d74e0677ce5179d9144daa91d659f606f28da2d4). 

I still kept the `PoWRateLimiter::cleanup_challenge_cache` method. Users that don't set the tokio flag would have to use this clone approach to implement any similar task for cleanup. The alternative would be for us to re-add an accessor method for the cache, but I think this way is fine.

crates/pow/src/lib.rs
```diff
@@ -49,22 +48,31 @@ pub struct PoWRateLimiterConfig {
...
+            loop {
+                interval.tick().await;
+                let current_time = SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .expect("current timestamp should be greater than unix epoch")
+                    .as_secs();
+                challenge_cache
+                    .write()
+                    .expect("challenge cache lock should not be poisoned")
+                    .drop_expired_challenges(current_time);
```
bobbinth: I would convert this into a small helper function to reduce code duplication - e.g., something like:
```rust
fn cleanup_challenge_cache(challenge_cache: Arc<RwLock<ChallengeCache>>) {
    let current_time = SystemTime::now()
        .duration_since(UNIX_EPOCH)
        .expect("current timestamp should be greater than unix epoch")
        .as_secs();
   challenge_cache
        .write()
        .expect("challenge cache lock should not be poisoned")
        .drop_expired_challenges(current_time);
}
```

PR #204, opened 2025-12-22
feat: add mint-script crate

### LINKED ISSUE #76: Move transaction script into a separate crate
From https://github.com/0xMiden/miden-faucet/pull/56#pullrequestreview-3171361734:
> In the longer run, we should put the transaction script MASM into a separate crate and make it appear as a "package". This is how eventually the compiler will be used to create transaction scripts in Rust (i.e., there will be a separate crate per transaction script).
>
> The same crate may contain some kind of "constructor" that could make building transactions using this script easier.

TomasArrachea:
Closes #76 

This PR creates a new `mint-tx` crate that implements the custom faucet transaction script in Rust. 

The `faucet` crate calls the compiler on runtime to generate the `.masp` and load it in the running faucet.


crates/faucet/src/lib.rs
```diff
@@ -140,10 +140,11 @@ impl Faucet {
         }
         client.set_setting(DEFAULT_ACCOUNT_ID_SETTING.to_owned(), account.id()).await?;
 
-        // Compile the mint tx script
+        // Compile the mint tx script from Rust via cargo-miden.
         let workspace_root = Path::new(env!("CARGO_MANIFEST_DIR"));
         let package = build_project_in_dir(&workspace_root.join("../contracts/mint-tx"), true)?;
         let program = package.unwrap_program();
+        // TODO: use client.code_builder()?
```
igamigo: 👍

crates/contracts/mint-tx/src/lib.rs
```diff
@@ -3,50 +3,15 @@
...
-    while i < bytes.len() {
-        result = result * 10 + (bytes[i] - b'0') as u64;
-        i += 1;
-    }
-    result
-}
-
 #[tx_script]
-fn run(arg: Word) {
+fn run(arg: Word, account: &mut Account) {
```
igamigo: I know this is in-progress, but would be great to add some doc comments to this explaining what type of advice data it expects, etc.

crates/faucet/src/lib.rs
TomasArrachea: IIUC, currently the code builder is intended to be used with MASM code. In this case, we already have a compiled Package generated by the compiler, so I'm not sure if using the code builder is needed. We are making some type conversions but eventually the protocol will simply expose a `TransactionScript::from_package`.

Anyway, It would be really convinient if the code builder was able to take rust code as input and use the compiler, though it would be a larger change.

TomasArrachea:
Closing because after https://github.com/0xMiden/faucet/pull/262 this won't be needed.

PR #239, opened 2026-04-16
feat: update to client v0.14.3

TomasArrachea:


bin/faucet/frontend/package.json
```diff
@@ -15,7 +15,7 @@
   "dependencies": {
     "@demox-labs/miden-wallet-adapter-base": "^0.10.0",
     "@demox-labs/miden-wallet-adapter-miden": "^0.10.0",
-    "@miden-sdk/miden-sdk": "^0.14.0"
+    "@miden-sdk/miden-sdk": "^0.14.2"
```
igamigo: Should be 0.14.3

igamigo (approved):

bin/faucet/frontend/package.json
TomasArrachea: Is that version published? On npm the latest seems to be v0.14.2: https://www.npmjs.com/package/@miden-sdk/miden-sdk

igamigo: I had the same problem where it did not show it to me, but going to the URL uncovers it: https://www.npmjs.com/package/@miden-sdk/miden-sdk/v/0.14.3

PR #241, opened 2026-04-21
fix: add store reference to sync process

igamigo:
The issue is that the faucet was constructing `StateSync` without giving it the local store. In `miden-client 0.14.3`, that store is needed for public account syncing (not for anything note or transaction related): the client reads the locally tracked account storage to know which map slots to request from the node. Without it, the client asks the node for account details but does not request any storage map details. The node response is then internally consistent from the node’s perspective: it includes the account storage header, which says `allowed_policy_proc_roots` is a map, but `map_details` is empty because the client did not ask for that map.

Devnet likely did not hit this because it was on a different client path/version (previous client version did not use to have this), or because the faucet account did not enter the specific out-of-sync state that triggers public account detail parsing. Older client versions fetched full public accounts differently and did not rely on `StateSync`’s optional store in this same way.

On the client side we will need to improve this as this is not ideal (at the very least it should be better signaled; but also this should not be 100% necessary) - will create a separate issue for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.14.2 (2026-04-21)
+
+- Fixed faucet state sync to request storage map details for tracked public accounts ([#241](https://github.com/0xMiden/faucet/pull/241)).
+
 ## 0.14.1 (2026-04-16)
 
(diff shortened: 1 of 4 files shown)


Mirko-von-Leipzig (approved):

Dominik1999 (approved):
Looks good to me

PR #243, opened 2026-04-21
feat: add note transport url to metadata response

### LINKED ISSUE #238: Expose the NTL endpoint used
The faucet accepts a `--note-transport-url` parameter to set the NTL. Once set, it is not visible for users which note transport layer is being used, so it would be useful to add it in the `/metadata` endpoint.

TomasArrachea:
Closes https://github.com/0xMiden/faucet/issues/238

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Changelog
 
-## Unreleased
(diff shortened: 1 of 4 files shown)


igamigo (approved):
LGTM!

PR #244, opened 2026-04-28
chore: update miden-client to v0.14.5

TomasArrachea:


SantiagoPittella (approved):
Looks good, do we need to update any note script in this repo?

TomasArrachea:
Closing this PR as the breaking changes of the release don't affect the faucet, so we don't need to update the version.

TomasArrachea:
We are making a patch release to improve observability, so we can also bump the client's patch version while at it.

PR #245, opened 2026-04-28
chore: impprove observability

SantiagoPittella:
Currently on testnet the faucet seems to be flaky, and in the trace we see that it errors when submitting transactions causing the users to receive an internal error. With the existing setup we did not have more information in the traces, make it difficult for us to debug this.

This PR adds more information to the traces, we may want to rollback this changes in the future.

### DIFF
--- a/crates/faucet/src/lib.rs
+++ b/crates/faucet/src/lib.rs
@@ -12,7 +12,7 @@ use miden_client::builder::ClientBuilder;
 use miden_client::crypto::{RandomCoin, Rpo256};
 use miden_client::keystore::{FilesystemKeyStore, Keystore};
 use miden_client::note::{Note, NoteAttachment, NoteError, NoteId, P2idNote};
-use miden_client::rpc::{Endpoint, GrpcClient, RpcError};
+use miden_client::rpc::{Endpoint, GrpcClient, GrpcError, RpcError};
 use miden_client::store::{NoteFilter, TransactionFilter};
 use miden_client::sync::{StateSync, StateSyncInput, SyncSummary};
 use miden_client::transaction::{
@@ -430,10 +430,11 @@ impl Faucet {
         skip_all,
         err,
         fields(
-            grpc.endpoint = tracing::field::Empty,
-            grpc.code = tracing::field::Empty,
-            grpc.endpoint_error = tracing::field::Empty,
-            rpc.error_kind = tracing::field::Empty,
+            rpc.system = tracing::field::Empty,
+            rpc.method = tracing::field::Empty,
+            rpc.grpc.status_code = tracing::field::Empty,
+            exception.type = tracing::field::Empty,
+            exception.message = tracing::field::Empty,
         )
     )]
     async fn submit_new_transaction(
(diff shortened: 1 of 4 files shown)


crates/faucet/src/lib.rs
```diff
@@ -279,7 +279,8 @@ impl Faucet {
                 Ok(()) => (),
                 Err(error) => {
                     if let Some(ClientError::RpcError(_)) = error.downcast_ref::<ClientError>() {
-                        error!(?error, "RPC error, discarding batch");
+                        let error_chain = format!("{error:#}");
```
TomasArrachea, line 282: Would be nice to check if this is actually showing the whole error chain

TomasArrachea (approved):
LGTM. Would be nice to check using jaegger that the traces are shown correctly.

SantiagoPittella:
View of the traces with this changes 
image

crates/faucet/src/lib.rs
SantiagoPittella: I'm not quite sure how to force the error here

SantiagoPittella:
This is the trace with a forced error in the node when submitting a proven tx

Screenshot 2026-04-28 at 18 07 36

crates/faucet/src/lib.rs
```diff
@@ -512,6 +567,26 @@ impl Faucet {
...
+fn record_grpc_error_fields(err: &ClientError) {
+    let span = tracing::Span::current();
+    if let ClientError::RpcError(rpc_err) = err {
+        span.record("rpc.error_kind", tracing::field::display(rpc_err));
+        if let RpcError::RequestError { endpoint, error_kind, endpoint_error, .. } = rpc_err {
+            span.record("grpc.endpoint", tracing::field::display(endpoint));
+            span.record("grpc.code", tracing::field::debug(error_kind));
+            if let Some(ee) = endpoint_error {
+                span.record("grpc.endpoint_error", tracing::field::display(ee));
+            }
```
Mirko-von-Leipzig: FYI there are open-telemetry standards for some of these (not that one has to follow them, but it does mean integration into backend tooling gets extra semantic meaning).

https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/

Mirko-von-Leipzig (commented):
Not a thorough review, my main input would be that if you want to follow open-telemetry specifications then one probably wants something a bit more general (but its a lot more work).

Might be something worth working towards in the future; though you would also have to consider this in the context of the client.

igamigo:
Until this gets deployed, is it a possibility to try to locate the faucet's requests in the node and try to get logs from there? Should be able to locate it through the account ID, no?

Mirko-von-Leipzig:
> Until this gets deployed, is it a possibility to try to locate the faucet's requests in the node and try to get logs from there? Should be able to locate it through the account ID, no?

They're currently not included in all areas. But I don't see any failed transaction submissions for example (like none at all). So I'm not entirely sure what's going on now..

crates/faucet/src/lib.rs
SantiagoPittella: Thanks, I wasn't aware of that. I updated at least the ones that I introduced in my PR

Mirko-von-Leipzig (approved):

PR #246, opened 2026-04-29
chore: bring main

TomasArrachea:
Merge without squash

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ license      = "MIT"
 readme       = "README.md"
 repository   = "https://github.com/0xMiden/miden-faucet"
 rust-version = "1.93"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -12,7 +12,7 @@ license      = "MIT"
 readme       = "README.md"
 repository   = "https://github.com/0xMiden/miden-faucet"
 rust-version = "1.93"
-version      = "0.14.2"
+version      = "0.14.3"
```
SantiagoPittella: are we going to use 0.15 here?

TomasArrachea: Updated it to v0.15.0

SantiagoPittella (approved):

PR #250, opened 2026-05-21
feat: update miden-client

TomasArrachea:
TODO
- ~Reexport AuthMethod from miden-client to avoid extra miden-standards dep.~


bin/faucet/Cargo.toml
```diff
@@ -21,6 +21,7 @@ miden-pow-rate-limiter = { features = ["tokio"], workspace = true }
 # Miden dependencies.
 miden-client              = { features = ["tonic"], workspace = true }
 miden-client-sqlite-store = { workspace = true }
+miden-standards           = { workspace = true }
```
SantiagoPittella: This was added to use `AuthMethod` , which is missing in the client right? Was that fixed on the client side already?

SantiagoPittella (approved):

bin/faucet/Cargo.toml
TomasArrachea: This will be exported after https://github.com/0xMiden/miden-client/pull/2229 gets merged.

PR #255, opened 2026-06-15
fix: Add asset callbacks

igamigo:
- Looks into acconut storage for deciding whether to activate callbacks
- Modifies test to involve this check
- Bumps patch version to release new binaries


Mirko-von-Leipzig (approved):

PhilippGackstatter (approved):
Looks good!

PR #256, opened 2026-06-16
Fix broken doc links to the renamed faucet repo

gabrielbosio:


TomasArrachea (approved):
Thanks! I think there are two more references that can be updated here: https://github.com/0xMiden/faucet/blob/ad061ee4a1e694ff6b7154409189fd0eb5b53e3e/.github/actions/debian/action.yml#L115-L117

gabrielbosio:
> Thanks! I think there are two more references that can be updated here:
> 
> https://github.com/0xMiden/faucet/blob/ad061ee4a1e694ff6b7154409189fd0eb5b53e3e/.github/actions/debian/action.yml#L115-L117

Yes, it seems those links are redirecting atm but I should change those to the new repo.

PR #259, opened 2026-07-17
Bump miden-client to 0.16.0-alpha.1

gabrielbosio:


Cargo.lock
TomasArrachea, line 1: Looks like there are still miden dependencies from v0.15, probably because we should also update the dev dependencies here:
https://github.com/0xMiden/faucet/blob/7f21dd9f5bacd0560be52e090956b663bff9a463/bin/faucet/Cargo.toml#L56-L57

igamigo (approved):
LGTM! I pushed a commit ([1a7fe0c](https://github.com/0xMiden/faucet/pull/259/commits/1a7fe0c9006b8e82c50fa3ad2251de5aebb656b7)) to upgrade to the newly released WebClient alpha version, and that also changes the node dependency to build the protobuf bindings at build-time to avoid having a hard dependency (addresses https://github.com/0xMiden/faucet/pull/259#discussion_r3615809825)

Cargo.lock
igamigo: I think this was not originally upgraded because the node version uses protocol `0.16.0-alpha.2` and client uses protocol `0.16.0-alpha.4`. This caused an incompatibility. To resolve this, I pushed a small commit that builds the protobuf bindings on build-time and use that to only have a dependency on the gRPC interface and not the crate.

Issue #261, opened 2026-07-25
Migrate faucet to using a network account


igamigo:
Originally discussed [here](https://github.com/0xMiden/protocol/issues/3373#issuecomment-5075570902).
The idea is that, to avoid modifying the standard account components, on genesis we can prefund a set of accounts to allow for the different services to work well. 

The faucet could be modified to be deployed as a network account which gets operated by a wallet which the UI controls. Basically, in terms of specific changes:

- The faucet backend now cares about 2 accounts: the network faucet, and the faucet operator (which is a local account)
- We should make it so that both the network account ID and operator accounts are provided at instantiation time. The main faucet client really only cares about the latter of these, since it will be the one that drives the network account
- The UI now collects requests with solved PoW challenges, and basically uses the wallet to emit one network note per each request and sends it to the network faucet (the request/network note). It's unclear to me whether a single network note can carry information for minting multiple asset notes (we'd have to look into the faucet standard) but it'd be ideal if it could, to avoid bloating the network with mint requests.
 
Some points:

- Since the minting process now really takes 2 transactions, there's more end-to-end latency
- It's not super clear whether we can easily provide the same flows after having minted the notes. For example, the network faucet will not be able to use the NTL to send private notes. I think we should be able to know all the details from the asset notes beforehand, so after minting we can decide to instantly provide a `NoteFile` that users/wallets can import. We could also poll the node from the frontend to see when the end asset note gets created
- Should we provide other operator transactions?

cc @bobbinth

bobbinth:
> It's not super clear whether we can easily provide the same flows after having minted the notes. For example, the network faucet will not be able to use the NTL to send private notes. I think we should be able to know all the details from the asset notes beforehand, so after minting we can decide to instantly provide a `NoteFile` that users/wallets can import. We could also poll the node from the frontend to see when the end asset note gets created

Could another option be that we still follow the network account so that we can detect once the `MINT` note has been consumed and the `P2ID` note is issued? From that point on, everything should work as it does now. Basically, the latency would increase (hopefully by no more than 3 - 5 seconds), but the overall functionality would remain largely the same.

igamigo:
> Could another option be that we still follow the network account so that we can detect once the `MINT` note has been consumed and the `P2ID` note is issued? From that point on, everything should work as it does now. Basically, the latency would increase (hopefully by no more than 3 - 5 seconds), but the overall functionality would remain largely the same.

Yeah, this should work.

[...]

PR #262, opened 2026-07-28
feat: migrate faucet to using a network account

ricomateo:
This PR partially addresses #261 

Introduces the following changes:
- The faucet is now a network account instead of a public account.
- The backend now batches mint notes to the faucet account, sending transactions from an operator account. When the faucet account consumes the mint notes, it produces the P2id notes.
- Changes the `--import` flag in the `init` command, so that it now requires the path to an **operator** account. This flag must be paired with the `--faucet-account-id` (which is the id of the network faucet account).
- Updates the `make clippy` target to fail whenever it catches errors.

### How to test it

Start a local network (from the `node` repository) with the following command.
**Important:** The node version must be `v0.16.0-alpha.3`, otherwise the faucet will fail to init.

```bash
# from the node repository
git checkout v0.16.0-alpha.3
make local-network-up COMPOSE_PROFILE_ARGS="--profile telemetry --profile monitor --profile note-transport"
```

Once the network is up and running, navigate to the `faucet` repository and checkout to this PR's branch

```bash
git fetch origin pull/262/head:migrate-faucet-network-account
git checkout migrate-faucet-network-account
```

Install the faucet binaries

```bash
make install-faucet
```

Initialize the faucet against devnet

```bash
miden-faucet init \
  --token-symbol TEST \
  --decimals 6 \
  --max-supply 100000000000000000 \
  --network localhost
```

Start the faucet

```bash
miden-faucet start \
  --network localhost \
  --note-transport-url http://localhost:57292
```

Navigate to http://localhost:8080 and request a mint, or request it via the faucet client with the following command

```bash
miden-faucet-client mint \
  --url http://localhost:8000 \
  --amount 100000000 \
  --target-account <TARGET_ACCOUNT> 
```


crates/faucet/src/lib.rs
```diff
@@ -78,7 +99,8 @@ pub struct Faucet {
     tx_prover: Arc<dyn TransactionProver>,
     issuance: watch::Sender<AssetAmount>,
     max_supply: AssetAmount,
-    script: TransactionScript,
+    p2id_notes: P2idNoteCache,
+    pub operator_account_id: AccountId,
```
TomasArrachea: I think this does not need to be public

crates/faucet/src/lib.rs
```diff
@@ -315,21 +408,41 @@ impl Faucet {
...
+
+        // The faucet's transaction only creates the MINT notes; the P2ID notes are minted later by
+        // the network, so they never land in the client store.
+        // They are cached here for `get_note` to serve.
+        {
+            let mut cache = self.p2id_notes.write().expect("p2id note cache is poisoned");
+            for note in p2id_notes {
+                cache.insert(note.id().to_hex(), CachedP2idNote { note, after_block_num });
+            }
+        }
```
TomasArrachea, line 492: We are collecting all the p2id notes in a cache that never gets evicted. This can be a problem since it grows unbounded in memory. We should look into storing this notes in the client's store (though I think this cannot be done without some changes on the client) or even adding a cache cleanup using the `after_block_num` as a reference.

crates/faucet/src/lib.rs
ricomateo: Fixed here [bd91265](https://github.com/0xMiden/faucet/pull/262/commits/bd9126570c7ab60d72c498d191317365ea027719)

crates/faucet/src/lib.rs
ricomateo: Done here [454ff72](https://github.com/0xMiden/faucet/pull/262/commits/454ff729e2e8c92532ea72d3c565e31d1be79e83)

README.md
TomasArrachea, line 1: The docker instructions should also be updated here.

crates/faucet/src/lib.rs
```diff
@@ -543,6 +663,20 @@ impl Faucet {
...
+    }
+
+    /// Returns the cached P2ID note, if it exists. Otherwise returns `None`.
+    pub fn get_p2id_note(&self, note_id: NoteId) -> Option<CachedP2idNote> {
+        self.p2id_notes
+            .read()
+            .expect("p2id note cache is poisoned")
+            .get(&note_id.to_hex())
+            .cloned()
+    }
```
TomasArrachea, line 700: I think this method is only used on a test so we could get away without it.

crates/faucet/src/lib.rs
```diff
@@ -315,21 +435,46 @@ impl Faucet {
...
+            let mut cache = self.p2id_notes.write().expect("p2id note cache is poisoned");
+            prune_stale_p2id_notes(&mut cache, after_block_num);
+            // Only private notes are cached
+            let private_notes = p2id_notes
+                .into_iter()
+                .filter(|note| matches!(note.metadata().note_type(), ProtocolNoteType::Private));
+            for note in private_notes {
+                cache.insert(note.id().to_hex(), CachedP2idNote { note, after_block_num });
+            }
+        }
```
TomasArrachea, line 492: We should add the notes to the cache after we submit the txs, so we at least know that they were successfully sent to the node.

TomasArrachea (commented):
Looks good! Left a few comments inline. Also: 
- We can now remove the `crates/faucet/asm/tx_scripts/mint.masm` and `crates/faucet/build.rs` files.
- We should update the [CLI docs](https://github.com/lambdaclass/miden-faucet/blob/b9037ead5a101532987c34340bd291184a74bd04/docs/src/getting-started/cli.md?plain=1#L113) and the shipped [.env](https://github.com/lambdaclass/miden-faucet/blob/b9037ead5a101532987c34340bd291184a74bd04/bin/faucet/.env) with the changes from the CLI variables.

crates/faucet/src/lib.rs
```diff
@@ -384,39 +529,12 @@ impl Faucet {
...
-        let advice_map = [(note_data_commitment, note_data)];
-
-        TransactionRequestBuilder::new()
-            .custom_script(self.script.clone())
-            .extend_advice_map(advice_map)
-            .expected_output_recipients(expected_output_recipients)
-            .script_arg(note_data_commitment)
-            .build()
+        let notes: Vec<Note> = notes.to_vec();
+        TransactionRequestBuilder::new().own_output_notes(notes).build()
```
TomasArrachea: We could add `.expiration_delta(10)` here. Previously the `mint.masm` implementation did it but now we can just set it here.

crates/faucet/src/lib.rs
ricomateo: Fixed here [bf28733](https://github.com/0xMiden/faucet/pull/262/commits/bf287336453d92ddadfc123e9e64965e363651c4)

README.md
ricomateo: Updated the docker instructions for the `init` command here [f77d9f9](https://github.com/0xMiden/faucet/pull/262/commits/f77d9f9d87982bb1fd5ec353f92ab98f082cc1dc)

crates/faucet/src/lib.rs
ricomateo: Done here [357596b](https://github.com/0xMiden/faucet/pull/262/commits/357596baffd7ac9cf0ea4bedf6d0127c7d790047)

crates/faucet/src/lib.rs
ricomateo: This method exists because `Faucet`'s `p2id_notes` field is private, so removing it would mean making the field public so that is accessible from the test. I think it is better to just keep the method, what do you think?

TomasArrachea: Yes, but there ia `p2id_notes()` method that exposes it. Anyway this is just a nit you feel free to disregard.

crates/faucet/src/lib.rs
```diff
@@ -539,7 +539,10 @@ impl Faucet {
     fn create_transaction(notes: &[Note]) -> Result<TransactionRequest, TransactionRequestError> {
         // Build the transaction
         let notes: Vec<Note> = notes.to_vec();
-        TransactionRequestBuilder::new().own_output_notes(notes).build()
+        TransactionRequestBuilder::new()
+            .own_output_notes(notes)
+            .expiration_delta(10)
```
TomasArrachea: nit: we can move the expiration delta into a constant.

TomasArrachea (approved):
LGTM

crates/faucet/src/lib.rs
ricomateo: Done here [3adcdaa](https://github.com/0xMiden/faucet/pull/262/commits/3adcdaa0979d6cea1e328e871575baa5f894d152)

crates/faucet/src/lib.rs
```diff
@@ -43,9 +75,13 @@ use crate::types::AssetAmount;
 
 const COMPONENT: &str = "miden-faucet-client";
 
-const TX_SCRIPT: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/assets/tx_scripts/mint.txs"));
 const KEYSTORE_PATH: &str = "keystore";
+/// How long a P2ID note is kept in the cache, in blocks, before it is pruned.
+const NOTE_RETENTION_BLOCKS: u32 = 100;
+/// How many blocks after its reference block a mint transaction stays valid for.
+const MINT_TX_EXPIRATION_DELTA: u16 = 10;
```
igamigo: This is for the transaction that sends the `MintNote`, right? If so, let's make this clearer onn the doc comment.

crates/faucet/src/lib.rs
```diff
@@ -96,18 +150,43 @@ pub struct FaucetConfig {
...
+    /// is submitted.
     pub async fn init(
         config: &FaucetConfig,
-        account: Account,
-        secret_key: &AuthSecretKey,
-        deploy: bool,
+        faucet_account: FaucetAccount,
+        operator_secret_key: &AuthSecretKey,
+        operator_account: Account,
     ) -> anyhow::Result<()> {
```
igamigo, line 185: For the existing account path, let's make sure the account exists by querying the node with a `GetAccount` request before proceeding.

crates/faucet/src/lib.rs
```diff
@@ -127,23 +206,58 @@ impl Faucet {
             Arc::new(GrpcClient::new(&config.node_endpoint, config.timeout.as_millis() as u64));
         let state_sync_component =
             StateSync::new(grpc_client.clone(), Arc::new(note_screener), None);
-        Self::sync_state(account.id(), &mut client, &state_sync_component).await?;
+        Self::sync_state(&[faucet_account_id], &mut client, &state_sync_component).await?;
```
igamigo: Let's sync using the operator account here.

crates/faucet/src/lib.rs
```diff
@@ -127,23 +206,58 @@ impl Faucet {
...
+        if matches!(faucet_account, FaucetAccount::Existing(_)) {
+            let faucet_account = client
+                .get_account(faucet_account_id)
+                .await
+                .context("failed to read the faucet account from the store")?
+                .with_context(|| format!("faucet account {faucet_account_id} is not tracked"))?;
 
-        let add_result = client.add_account(&account, false).await;
+            check_faucet_owner_matches_operator(&faucet_account, operator_account.id())?;
+        }
```
igamigo: nit: I think we should verify that the account owner matches the operator for all cases. I know in the "new account" case it seem unnecessary but it also doesn't hurt and we don't need to branch here

crates/faucet/src/lib.rs
```diff
@@ -384,38 +535,14 @@ impl Faucet {
...
-
+        let notes: Vec<Note> = notes.to_vec();
         TransactionRequestBuilder::new()
-            .custom_script(self.script.clone())
-            .extend_advice_map(advice_map)
-            .expected_output_recipients(expected_output_recipients)
-            .script_arg(note_data_commitment)
+            .own_output_notes(notes)
+            .expiration_delta(MINT_TX_EXPIRATION_DELTA)
             .build()
```
igamigo, line 559: Now that I think about it, here we could use something like the `expected_future_notes()` setter. 
This setter adds notes to the client's notes table, where you know it will be created as part of a future transaction. In this case, for example, we know there will be an asset note created not as part of this transaction, but rather as an output of the network transaction. However, we only have this for input notes, but we'd want this to be added to the output notes table. Let's add an issue on `rust-sdk` to tackle this

crates/faucet/src/lib.rs
```diff
@@ -653,43 +837,201 @@ fn build_p2id_notes(
...
+
+        let storage = match p2id_note.metadata().note_type() {
+            ProtocolNoteType::Public => {
+                MintNoteStorage::new_fungible_public(recipient, asset, tag)?
+            },
+            ProtocolNoteType::Private => {
+                MintNoteStorage::new_fungible_private(recipient.digest(), asset, tag)
+            },
+        };
+        // SAFETY: `faucet_id` is a public (network) account
```
igamigo, line 874: Is this checked anywhere?

crates/faucet/src/lib.rs
```diff
@@ -653,43 +837,201 @@ fn build_p2id_notes(
...
+
+/// Reads the ID of the faucet whose asset the chain charges fees in from the genesis block header.
+pub async fn fetch_fee_faucet_id(
+    node_endpoint: &Endpoint,
+    timeout: Duration,
+) -> anyhow::Result<AccountId> {
+    let (genesis, _) = GrpcClient::new(node_endpoint, timeout.as_millis() as u64)
+        .get_block_header_by_number(Some(BlockNumber::GENESIS), false)
+        .await
+        .context("failed to fetch the genesis block header")?;
```
igamigo, line 897: This is fine for now, but ideally we would read from the latest block available, which we should already have on the client. Let's add a function in `miden-client` to read the latest stored block header.

igamigo (approved):
LGTM!

crates/faucet/src/lib.rs
ricomateo: Done here [e223e21](https://github.com/0xMiden/faucet/pull/262/commits/e223e2105fb7b4dae78511e446e7bb908de8454e)

crates/faucet/src/lib.rs
ricomateo: Done here [ff214a5](https://github.com/0xMiden/faucet/pull/262/commits/ff214a5ca28c500892d82b90def2c3999045b1f6)

crates/faucet/src/lib.rs
ricomateo: Done here [0a7d01f](https://github.com/0xMiden/faucet/pull/262/commits/0a7d01fa1439e35d811ab43b975d4a7db67c59b6)

crates/faucet/src/lib.rs
ricomateo: Done here [4112e4a](https://github.com/0xMiden/faucet/pull/262/commits/4112e4ab9d174cc0eb50813f4095b58731e8a5ed)

crates/faucet/src/lib.rs
ricomateo: When importing a faucet account, this is now checked, after the change introduced at https://github.com/0xMiden/faucet/pull/262/changes/ff214a5ca28c500892d82b90def2c3999045b1f6 (if the account is private, that check fails with `faucet account has no public state on the node`).

When creating a new faucet account there is no check, however there is no need to check that, since `create_network_faucet_account` creates the account using `NetworkAccount::builder`, which sets `AccountType::Public`.

crates/faucet/src/lib.rs
ricomateo: I just created an issue on `rust-sdk` https://github.com/0xMiden/rust-sdk/issues/2395

crates/faucet/src/lib.rs
ricomateo: I just opened a PR adding a method for this https://github.com/0xMiden/rust-sdk/pull/2397

igamigo (approved):

PR #263, opened 2026-07-29
chore: point clients at the `encrypted-tx-inputs` branches

igamigo:
NOTE: This ships the vendored JS package built from https://github.com/0xMiden/web-sdk/pull/252, but we might want to release `miden-sdk` to NPM before shipping this

### DIFF
--- a/bin/faucet/Cargo.toml
+++ b/bin/faucet/Cargo.toml
@@ -65,7 +65,7 @@ uuid          = { workspace = true }
 # Used to generate the node RPC types for the stub node used in tests 
 # (so the dependency only lies in the protobuf interface).
 [build-dependencies]
-miden-node-proto-build = { version = "0.16.0-alpha.2" }
+miden-node-proto-build = { git = "https://github.com/0xMiden/node.git", rev = "b4ece8c9036745840e1cdd904a10b573283c37e2" }
 tonic-prost-build      = { version = "0.14" }
 
 # Required to avoid false positives in cargo-machete
--- a/bin/faucet/Cargo.toml
+++ b/bin/faucet/Cargo.toml
@@ -51,16 +51,17 @@ tracing-subscriber    = { workspace = true }
 url                   = { workspace = true }
 
 [dev-dependencies]
-fantoccini    = { version = "0.22" }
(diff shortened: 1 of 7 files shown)


bin/faucet/frontend/package.json
```diff
@@ -15,7 +15,7 @@
   "dependencies": {
     "@demox-labs/miden-wallet-adapter-base": "^0.10.0",
     "@demox-labs/miden-wallet-adapter-miden": "^0.10.0",
-    "@miden-sdk/miden-sdk": "^0.16.0-alpha.1"
+    "@miden-sdk/miden-sdk": "file:./miden-sdk-miden-sdk-0.16.0-alpha.1.tgz"
```
juan518munoz, line 18: I might be reading this json wrong, but this looks like a local dependency, we should fix it once the miden-sdk package gets a proper release.

bin/faucet/Cargo.toml
```diff
@@ -51,16 +51,17 @@ tracing-subscriber    = { workspace = true }
...
-prost         = { version = "0.14" }
-reqwest       = { features = ["json"], workspace = true }
-serde_json    = { workspace = true }
-tokio         = { features = ["macros", "process"], workspace = true }
-tonic-prost   = { version = "0.14" }
-tonic-web     = { version = "0.14" }
-uuid          = { workspace = true }
+fantoccini     = { version = "0.22" }
+miden-client   = { features = ["testing"], workspace = true }
+miden-protocol = { version = "0.16.0-alpha.4" }
```
juan518munoz: We could probably re-export `SigningKey` from the client to prevent having this extra dependency.

juan518munoz (approved):
Looks good, I would hold on merging until we have the necessary `miden-sdk` release so we don't push the generated binary.

bin/faucet/frontend/package.json
igamigo: Yeah, as mentioned in the description we'll need to do something like that until we can reference an NPM version (not sure we will soon, but I'll open an issue to address this otherwise):

> NOTE: This ships the vendored JS package built from https://github.com/0xMiden/web-sdk/pull/252, but we might want to release miden-sdk to NPM before shipping this

Issue #254, opened 2026-06-13
Address package/docker image publishing


igamigo:
The publish workflow publishes a docker image and outputs Debian packages. 
There are a couple of things to note here:

- The docker workflow seems to be failing potentially due to stale permissions: https://github.com/0xMiden/faucet/actions/runs/27452179614
- We probably don't need both Debian packages and Docker images if only one of these mechanisms is used for deploying to different network environments (although not sure which one is currently being used cc @Mirko-von-Leipzig)

Mirko-von-Leipzig:
Only docker is required. We will be removing the node debian packages hopefully soon.

igamigo:
We can create a PR for removing the Debian packages, but I believe the problem with the current Docker publishing flow is that this repo has no access to the [existing package](https://github.com/0xmiden/node/pkgs/container/miden-faucet). An org admin should add the correct role, or transfer the package repository from `node` to `faucet` cc @Mirko-von-Leipzig .
As part of this we should also update the Github Actions versions.

Mirko-von-Leipzig:
@igamigo sorry that took a while; didn't realize I had permissions to assign this repo.

This repo has admin permissions for the `miden-faucet` package you linked. Could we get this done with arm + amd support? It would also be good to have something like https://github.com/0xMiden/node/pull/2419 so we can publish arbitrary commits (e.g. for demo's).

Once complete we can drop the node performing this duty.

gabrielbosio:
Confirming that now the Docker publish workflow succeeds. I just ran the workflow for v0.15.0 to test: https://github.com/0xMiden/faucet/actions/runs/30845750754/job/91793492457

gabrielbosio:
Now that PRs that address this issue got merged, I tested 2 cases of the manual Docker publish workflow, both succeeded:
- Dry run: https://github.com/0xMiden/faucet/actions/runs/30929857582
- Publish: https://github.com/0xMiden/faucet/actions/runs/30936554477

The workflow publishes the container using two tags:
- One mutable (The one passed as input): https://github.com/0xmiden/node/pkgs/container/miden-faucet/1098732475?tag=test
- One Immutable (Following the form `sha-<commit12>-<runid>-<attempt>`): https://github.com/0xmiden/node/pkgs/container/miden-faucet/1098732475?tag=sha-4bcfa0b79715-30936554477-1

Closing this issue.

PR #265, opened 2026-08-03
Publish multi-arch Docker images and enable manual per-commit builds

gabrielbosio:
Adds support for both arm64 and amd64 architectures in the Docker publish workflow. It also makes the manual workflow to receive commit SHAs instead of release tags.

Additional changes:

- Given the scenario where a run starts executing when a run has already started (e.g. a release run and a manual run), this PR makes both runs to be executed in sequence. In this case, the first run will start executing once the second run finishes.
- Reduces permissions to be content read-only.
- Adds a preflight job that checks whether the commit SHA input is from a valid commit.
- Fixes docs so they don't point to `latest` Docker images since they could never exist in the repository.
- Because the Docker publish workflow already includes building Docker images for both arm64 and amd64 architectures, the Docker build workflow gets obsolete with this change. This PR addresses this by removing that workflow and adding the necessary logic in the Docker publish workflow so that the latter gets executed as a dry run on every PR and push to `next`.

Part of #254. This PR along with #266 close the related issue.


.github/workflows/docker.yml
```diff
@@ -6,59 +6,217 @@ on:
 
   workflow_dispatch:
     inputs:
-      version:
-        description: "Version to release (E.G. v0.10.0-rc.1, v0.10.0). Corresponding git tag must already exist."
+      commit:
+        description: "Full git commit SHA to build (40 lowercase hex characters)."
         required: true
         type: string
+      tag:
```
Mirko-von-Leipzig, line 27: What is `type` for?

.github/workflows/docker.yml
Mirko-von-Leipzig, line 1: If you add a `dryrun` | `publish` flag input to this, then you can also remove the `docker-build.yml` workflow by using this one in `dryrun` mode.

Mirko-von-Leipzig (commented):
Code LGTM, but leaving approval for someone else

.github/workflows/docker.yml
Mirko-von-Leipzig: https://github.com/0xMiden/node/pull/2419

.github/workflows/docker.yml
gabrielbosio: It's required and it tells Github how to show the input in the "Run workflow" dropdown. For instance, `type: string` means the input will be entered through a single-line text field.

.github/workflows/docker.yml
gabrielbosio: Done in [87537d9](https://github.com/0xMiden/faucet/pull/265/commits/87537d967c6c517b5ea815c05454797626ff6116)

igamigo (approved):
LGTM

PR #268, opened 2026-08-04
fix: log the full error chain in the mint span

### LINKED ISSUE #258: Sync state does not log source errors
On a failing `sync_state()`, the innermost error we are logging is ["failed to sync state"](https://github.com/0xMiden/faucet/blob/main/crates/faucet/src/lib.rs#L247). We should make it so we provide a better look into what went wrong.

ricomateo:
Closes https://github.com/0xMiden/faucet/issues/258

### Motivation

When `sync_state` fails, the emitted error only includes the message `"failed to sync state"`, dropping the chain of errors. This makes the traces hard to debug. For example, this is how the error looks when inspecting the traces in Jaeger:

log_before_

This is caused because the `mint` function is instrumented with the `#[instrument(err)]` attribute, which by default it records the error through its `Display` implementation. For `anyhow::Error`, `Display` prints only the outermost context (the `caused by` chain is dropped). So an error inside the mint path is recorded including only the outermost context (e.g. `"failed to sync state"`), missing the actual cause.
The same happens for the `submit_new_transaction` path in the `mint` function.

### Changes

Changes the `mint` function instrumentation attribute from `#[instrument(err)]` to `#[instrument(err(Debug))]` so that the error is recorded using its `Debug` implementation, which includes the full error chain.

With this change, the error now contains the actual cause:

log_now_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.16.0 (TBD)
 
-- Emit the full error chain in the `faucet.mint` span by recording errors with their `Debug` implementation ([#267](https://github.com/0xMiden/faucet/pull/267)).
+- Improved mint failure observability: emit the full error chain in the `faucet.mint` span by recording errors with their `Debug` implementation ([#267](https://github.com/0xMiden/faucet/pull/267)).
 
 ## 0.16.0-alpha.1 (2026-07-20)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.16.0 (TBD)
+
+- Emit the full error chain in the `faucet.mint` span by recording errors with their `Debug` implementation ([#267](https://github.com/0xMiden/faucet/pull/267)).
+
 ## 0.16.0-alpha.1 (2026-07-20)
 
 - Updated miden-client dependency to v0.16.0-alpha.1 ([#259](https://github.com/0xMiden/faucet/pull/259))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.16.0 (TBD)
+
+- Emit the full error chain in the `faucet.mint` span by recording errors with their `Debug` implementation ([#267](https://github.com/0xMiden/faucet/pull/267)).
```
juan518munoz: minor nit (feel free to disregard): we could prefix the entry like we do [here](https://github.com/0xMiden/faucet/pull/268/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR23):

```suggestion
- Improved mint failure observability: emit the full error chain in the `faucet.mint` span by recording errors with their `Debug` implementation ([#267](https://github.com/0xMiden/faucet/pull/267)).
```

juan518munoz (approved):

CHANGELOG.md
ricomateo: Nice catch!

igamigo (approved):
LGTM! One other thing I thought of: now that the faucet will be a network faucet, we should emit events for what notes were committed and consumed. For example, it would be great to be able to see the amount of time that passes between the time in which the operator submits the tx, then the note gets committed, and finally it gets picked up and consumed by the ntx builder. We should have this information as part of the `SyncSummary` that we get from doing `Client::sync_state`

PR #288, opened 2026-08-27
feat: improve frontend loading times

TomasArrachea:
- Improve frontend loading times: now SDK WASM download is deferred until the page finishes loading and is now compressed using brotli, the JS bundle is minified, the header image was converted to lossless WebP (207KB -> 70KB), and all static assets are served with cache headers.
- Fixed the issuance counter staying empty until the next mint: the initial SSE issuance event raced the metadata fetch and was dropped; the value is now cached and rendered once the metadata arrives.
- Added loading spinners for the footer values (faucet address, tokens claimed) and disabled the token amount selector until its options load.


igamigo (approved):
LGTM! Thanks

PR #289, opened 2026-08-28
chore: fee support for faucet

igamigo:


docs/src/getting-started/quick-start.md
```diff
@@ -80,16 +80,25 @@ miden-faucet start --network localhost
 
 ### Development
 
-Connect to the node deployed in Miden Devnet.
+Connect to the node deployed in Miden Devnet. Devnet charges transaction fees
+(`verification_base_fee = 10000`) in its native `MIDEN` asset, which is issued by the genesis network
+faucet `0x8e085366ccd7cfb112967733836898`. On such a chain a new faucet account cannot be created
+(it could not pay for its own deployment), so the faucet is initialized against the genesis faucet
+with the operator account file that owns it:
```
igamigo: Let's avoid mentioning specific base fees, genesis commitments, etc here and make this more general

docs/src/getting-started/cli.md
```diff
@@ -65,6 +65,29 @@ miden-faucet start \
 A newly created faucet account is always deployed by submitting an empty transaction. An imported
 faucet account is already on-chain, so nothing is submitted.
 
+### Fee-charging chains
+
+On a chain whose genesis sets a non-zero `verification_base_fee`, every transaction pays a fee in the
+chain's native asset from the executing account's vault. The kernel charges
+`verification_base_fee * (ilog2(execution cycles) + 1)`, so a transaction costs a small multiple of
+the base fee and never more than `30 * verification_base_fee` (the maximum number of execution
+cycles is `2^29`).
```
igamigo: I wouldn't be too specific here either, just mentioning that the faucet operator needs to have some tokens for transactions should be fine

crates/faucet/src/lib.rs
```diff
@@ -130,12 +157,15 @@ pub struct CachedP2idNote {
...
     client: Client<FilesystemKeyStore>,
+    store: Arc<dyn Store>,
     state_sync_component: StateSync,
     tx_prover: Arc<dyn TransactionProver>,
     issuance: watch::Sender<AssetAmount>,
     max_supply: AssetAmount,
     p2id_notes: P2idNoteCache,
     operator_account_id: AccountId,
+    /// The chain's fee parameters, read from the genesis block header.
+    fee_parameters: FeeParameters,
```
igamigo: Let's remove this and take them through the client's latest block header (we should be able to get it through `client` directly)

crates/faucet/src/lib.rs
```diff
@@ -130,12 +157,15 @@ pub struct CachedP2idNote {
 pub struct Faucet {
     id: FaucetId,
     client: Client<FilesystemKeyStore>,
+    store: Arc<dyn Store>,
```
igamigo: Same here, let's remove this field

crates/faucet/src/lib.rs
```diff
@@ -810,6 +928,161 @@ impl Faucet {
...
+
+/// Reads the chain's fee parameters from the genesis block header stored by the client.
+async fn read_fee_parameters(client: &Client<FilesystemKeyStore>) -> anyhow::Result<FeeParameters> {
+    let (genesis, _) = client
+        .get_block_header_by_num(BlockNumber::GENESIS)
+        .await
+        .context("failed to read the genesis block header from the store")?
+        .context("the genesis block header is not in the store")?;
+    Ok(genesis.fee_parameters().clone())
+}
```
igamigo, line 891: Instead of the genesis block, let's get the latest block we have

crates/faucet/src/lib.rs
```diff
@@ -810,6 +928,161 @@ impl Faucet {
...
+/// A new faucet account is deployed by an empty transaction that its own auth procedure has to pay
+/// for. Its vault is empty and, being a network account without a wallet component, it cannot
+/// receive the fee asset, so on a fee-charging chain such a deployment can never succeed.
+fn ensure_new_faucet_can_be_deployed(fee_parameters: &FeeParameters) -> anyhow::Result<()> {
+    anyhow::ensure!(
+        fee_parameters.verification_base_fee() == 0,
+        "cannot create a new faucet account on a chain that charges transaction fees \
+         (verification base fee {}): the account would have to pay for its own deployment out of \
+         an empty vault. Import an existing faucet with `--import` and `--faucet-account-id` \
+         instead",
```
igamigo: Technically we could deploy a new account if the transaction consumes a note with the correct asset amount. But I wouldn't do anything about this now, only maybe adjust the wording a little bit

crates/faucet/src/lib.rs
```diff
@@ -810,6 +928,161 @@ impl Faucet {
...
+            operator.account.id = %operator_account.id(),
+            fee.faucet.id = %fee_parameters.fee_faucet_id(),
+            fee.verification_base_fee = fee_parameters.verification_base_fee(),
+            fee.max_per_transaction = max_transaction_fee(fee_parameters),
+            operator.fee_asset_balance = fee_asset_balance(operator_account, fee_parameters),
+        },
+        "The chain charges transaction fees: the operator account must hold the fee asset before \
+         the faucet can mint",
+    );
+}
```
igamigo: I would remove this, as it's probably not too useful on its own. Errors will report the required fee anyway

crates/faucet/src/lib.rs
```diff
@@ -810,6 +928,161 @@ impl Faucet {
...
+    if balance < required {
+        return Err(MintError::OperatorFeeBalanceTooLow {
+            operator_account_id: operator.id(),
+            fee_faucet_id: fee_parameters.fee_faucet_id(),
+            balance,
+            required,
+        });
+    }
+    Ok(())
+}
```
igamigo: I think we can remove all this. We should not have to calculate fees ourselves. Instead we can just execute and try to catch execute errors when we don't have enough assets to call `pay_fee`

crates/faucet/src/lib.rs
```diff
@@ -810,6 +928,161 @@ impl Faucet {
...
+
+/// Returns the amount of the chain's fee asset held in `account`'s vault, in base units.
+pub fn fee_asset_balance(account: &Account, fee_parameters: &FeeParameters) -> u64 {
+    let fee_asset_id = AssetId::new_fungible(fee_parameters.fee_faucet_id());
+    account
+        .vault()
+        .get_balance(fee_asset_id)
+        .map(|amount| amount.as_u64())
+        .expect("a fungible asset id should have a fungible composition")
+}
```
igamigo: We could use `AccountReader` instead of `Account` here (but as I mentioned in the other comment, we may not need this code)

crates/faucet/src/lib.rs
```diff
@@ -810,6 +928,161 @@ impl Faucet {
...
+) -> Option<FeeConversionInfo> {
+    if fee_parameters.verification_base_fee() == 0 {
+        return None;
+    }
+    let reads_conversion_info =
+        AccountComponentInterface::from_procedures(account.code().procedures())
+            .iter()
+            .any(|component| matches!(component, AccountComponentInterface::AuthSingleSig));
+    reads_conversion_info.then(|| FeeConversionInfo::one_to_one(fee_parameters.fee_faucet_id()))
+}
```
igamigo: I think we can also remove all this and basically hardcode it all for now

crates/faucet/src/lib.rs
```diff
@@ -810,6 +928,161 @@ impl Faucet {
...
+
+/// Declares the faucet as a foreign account of the operator's MINT transaction.
+///
+/// Creating a note with a `NetworkAccountTarget` attachment prices the note through an FPI into
+/// the target's fee policy, which reads the faucet's storage maps. Requesting every entry of every
+/// map slot up front serves the whole FPI from a single RPC call anchored at the transaction's
+/// reference block. It also keeps the request wire-compatible with node `0.16.0-rc.2`, which no
+/// longer answers per-key storage map requests in the shape this client version decodes.
+pub fn faucet_foreign_account(
+    faucet_account: &Account,
```
igamigo, line 931: I think we only need the `AccountHeader` here

crates/faucet/src/lib.rs
```diff
@@ -810,6 +928,161 @@ impl Faucet {
...
+/// The faucet's own sync never uses the store's tags, this keeps the store clean for anything else
+/// that reads it.
+async fn remove_tx_fee_note_tags(store: &dyn Store) -> Result<(), ClientError> {
+    for record in store.get_note_tags().await? {
+        if record.tag == TxFeeNote::TAG {
+            store.remove_note_tag(record).await?;
+        }
+    }
+    Ok(())
+}
```
igamigo: I think this is the case because it can eventually get reclaimed by the executor, no? If so, we may keep this for now but let's add an issue in the client repo for this functionality

igamigo (commented):
(Opened the PR to start the review process)

BTW with this, let's bump the crate versions to prepare a release

docs/src/getting-started/cli.md
juan518munoz: Done in af9d026a1a8c

```md
On a chain that charges transaction fees, every transaction pays in the chain's native asset out of
the executing account's vault, so the faucet operator account needs to hold enough of that asset to
cover the transactions the faucet submits and has to be topped up as its balance drains.
```

crates/faucet/src/lib.rs
juan518munoz: Removed in af9d026a1a8c

crates/faucet/src/lib.rs
juan518munoz: Removed in af9d026a1a8c

crates/faucet/src/lib.rs
juan518munoz: Harcoded as of af9d026a1a8c

crates/faucet/src/lib.rs
juan518munoz: Done in af9d026a1a8c

We still use this function on `init` to log if the account not holding the needed fee assets.

crates/faucet/src/lib.rs
juan518munoz: Done in af9d026a1a8c

crates/faucet/src/lib.rs
juan518munoz: Removed in af9d026a1a8c

crates/faucet/src/lib.rs
juan518munoz: Removed in af9d026a1a8c

crates/faucet/src/lib.rs
juan518munoz: Addressing [other comment](https://github.com/0xMiden/faucet/pull/289/changes#r3882838321) meant dropping this function.

af9d026a1a8c

crates/faucet/src/lib.rs
juan518munoz: We need `Account` to retrieve map slots, `AccountHeader` does not provide that. We may be able to cache the slot names at start up, then we can change the struct.

Does that sound ok?

crates/faucet/src/lib.rs
juan518munoz: Done in af9d026a1a8c

```rs
        "cannot create a new faucet account on a chain that charges transaction fees \
         (verification base fee {}): the faucet deploys it with an empty transaction, which has \
         to pay for itself out of the new account's empty vault. Import an existing faucet with \
         `--import` and `--faucet-account-id` instead",
```

juan518munoz (approved):

README.md
```diff
@@ -47,6 +47,43 @@ miden-faucet start \
   --network testnet
 ```
 
+## Running on a chain that charges fees (devnet)
+
```
TomasArrachea, line 51: I'd summarize this a bit. Important point is that the operator should be an account that is funded to pay for the fees.

TomasArrachea (approved):
LGTM

crates/faucet/src/lib.rs
igamigo: Ah, right! I missed that. I think we can leave it like that for now.

PR #292, opened 2026-09-02
feat: fund the operator account when its balance is below a threshold

### LINKED ISSUE #286: Ensure operator does not run out of fees
We need to have a mechanism to fund the operator account in case it runs out of money. Because the testnet/devnet oeprators will have funds from the genesis event, this is probably low priority.
Because the faucet uses the client, it should be simple to sync and keep P2ID notes and consume them every now and then. We could even consume them on transactions that mint notes.

ricomateo:
The faucet now funds the operator when its balance is below a threshold.

When executing a mint batch, the faucet includes an extra MINT note to fund the operator when its balance is below a threshold. When sending the mint transaction, the faucet now consumes any P2ID (or P2IDE) notes targeted to the operator, so the funds actually land in the account.

Closes #286


crates/faucet/src/lib.rs
```diff
@@ -1262,27 +1368,78 @@ pub fn create_faucet_operator_account() -> anyhow::Result<(Account, AuthSecretKe
...
+/// height has passed. A P2ID note carries no such condition.
+pub(crate) fn is_consumable_at(details: &NoteDetails, block_num: BlockNumber) -> bool {
+    let recipient = details.recipient();
+    if recipient.script().root() == P2idNote::script_root() {
+        return true;
+    }
+
+    P2ideNoteStorage::try_from(recipient.storage().items())
+        .is_ok_and(|storage| storage.timelock_height().is_none_or(|height| height <= block_num))
+}
```
igamigo: To simplify for now, I think we can just detect P2ID notes (which is what the faucet emits anyway)

igamigo: This will simplify detection as P2ID notes can be filtered by script root, and we can check that the operator can consume it by looking at the note storage.

crates/faucet/src/lib.rs
```diff
@@ -881,6 +927,66 @@ impl Faucet {
         );
         Ok(())
     }
+
+    /// Returns the P2ID notes that fund the operator.
+    async fn get_notes_targeted_to_operator(
+        &self,
+        block_num: BlockNumber,
+    ) -> anyhow::Result<Vec<Note>> {
```
igamigo: For safety, let's try to filter for P2ID notes that only carry the native asset (maybe exclusively). Let's also consume the ones that have the most assets first, and let's take a limited amount (say, 5-10). The reason is that there is a limit to how many notes we can consume in a single transaction and we don't want to unnecessarily prove a transaction that consumes a lot of notes that don't carry a lot of assets anyway.

igamigo (commented):
Looks great! Not a deep review, but leaving some general comments on things we could improve

crates/faucet/src/lib.rs
ricomateo: Removed all the logic to detect P2IDE notes here [1282a8a](https://github.com/0xMiden/faucet/pull/292/commits/1282a8a0b6b1b036c5a8fcc319b051b3e64af945).

crates/faucet/src/lib.rs
ricomateo: Changed `get_notes_targeted_to_operator` to cap the amount of notes to 5 and sort them descending by asset amount [9c0d778](https://github.com/0xMiden/faucet/pull/292/commits/9c0d7789797e0a9eb0cbb518db81c8b85d506ba4).

The notes are already being filtered by the asset (only the ones that carry the native asset are kept).
I renamed some helpers and variables (from `fee_faucet_id` to `native_fee_faucet_id`) here [4fdb9e6](https://github.com/0xMiden/faucet/pull/292/commits/4fdb9e67296a0987d2e46a75bd1814547d09050e) to make this more explicit.

crates/faucet/src/lib.rs
```diff
@@ -1262,6 +1374,34 @@ pub fn create_faucet_operator_account() -> anyhow::Result<(Account, AuthSecretKe
...
 }
 
+/// The amount of fungible assets a note carries.
+fn asset_amount(note: &Note) -> u64 {
+    note.assets().iter_fungible().map(|asset| asset.amount().as_u64()).sum()
+}
+
+/// Whether `details` describes a P2ID note payable to `target` that carries `fee_faucet_id`'s
+/// asset and nothing else.
+pub(crate) fn is_p2id_note_payable_to(
```
juan518munoz, line 1401: we should also check in this function (or somewhere else) that the note we are about to consume originated from the faucet, otherwise we would be open to consuming malicious notes.

crates/faucet/src/lib.rs
```diff
@@ -1374,6 +1515,68 @@ mod tests {
...
+    async fn mint_funds_the_operator_when_its_balance_is_low() {
+        let store = Arc::new(
+            SqliteStore::new(temp_dir().join(format!("{}.sqlite3", Uuid::new_v4())))
+                .await
+                .unwrap(),
+        );
+        // Set the initial operator balance below the threshold so that the next mint request
+        // triggers the funding mechanism
+        let initial_operator_balance = OPERATOR_FUNDING_THRESHOLD - 1;
+        let verification_base_fee = 0;
```
juan518munoz: shouldn't we test against a fee chargin chain?

crates/faucet/src/lib.rs
```diff
@@ -1542,20 +1758,26 @@ mod tests {
...
+        build_faucet_on_chain(store, 0, 0, false).await.0
     }
 
     /// Builds a faucet using a mock client on a chain charging `verification_base_fee`, with the
     /// operator holding `operator_fee_balance` base units of the chain's fee asset at genesis.
+    ///
+    /// `faucet_is_chain_fee_faucet` makes the chain charge fees in the faucet's own asset, the only
+    /// arrangement in which the faucet can fund its operator.
+    ///
+    /// Returns a tuple containing the faucet, the RPC api, and a funded account.
```
juan518munoz: This doesn't align with the actual return value.

crates/faucet/src/lib.rs
TomasArrachea, line 633: We are currently setting `self.funding_request_in_flight = true` when creating the p2id note, but in case the tx submission fails, we need to set it to false. That way the faucet will retry the funding.

crates/faucet/src/lib.rs
```diff
@@ -439,6 +456,15 @@ impl Faucet {
         let uncommitted_transactions =
             client.get_transactions(TransactionFilter::Uncommitted).await?;
 
+        // The node matches note inclusions by tag, and skips the query altogether when no tag is
+        // given. `add_account` registers a tag per tracked account, so passing them makes the sync
+        // return the P2ID notes payable to the operator.
+        let note_tags =
+            client.get_note_tags().await?.into_iter().map(|record| record.tag).collect();
```
TomasArrachea: Here we are passing the tags for both the faucet and the operator accounts, but we only need the operator one IIUC. This should not be a problem since the MINT notes directed to the faucet don't set the tag, but we might still want to remove it just in case.

TomasArrachea (approved):

crates/faucet/src/lib.rs
igamigo: I think this is fine. We should allow notes from other accounts to be sent to the faucet operator. Because we know it's P2ID notes, there's no room for allowing anything malicious. We should check that it contains native tokens as I mentioned in one of the other comments, though

igamigo (approved):
LGTM! Thanks! I think we should improve logging for this functionality. Would be great to report when a P2ID note was detected on a sync, and when a transaction consumes them.

crates/faucet/src/lib.rs
ricomateo: Done here [db1e849](https://github.com/0xMiden/faucet/pull/292/commits/db1e84915e63bc5742e9b0739052614d139ec07a)

crates/faucet/src/lib.rs
ricomateo: Done here [4d5ad2a](https://github.com/0xMiden/faucet/pull/292/commits/4d5ad2ae5f690f37ad17e656be0c3b2eeff2f3b2)

crates/faucet/src/lib.rs
ricomateo: Fixed here [dd25b29](https://github.com/0xMiden/faucet/pull/292/commits/dd25b295ed5e9fcda0b31d9a9e8e63b6bc644af4)

crates/faucet/src/lib.rs
```diff
@@ -457,10 +460,15 @@ impl Faucet {
...
-            client.get_note_tags().await?.into_iter().map(|record| record.tag).collect();
+        // given. `add_account` registers a tag per tracked account, so passing the operator note
+        // tag makes the sync return the P2ID notes payable to the operator
+        let note_tags = client
+            .get_note_tags()
+            .await?
+            .into_iter()
+            .filter(|record| record.source == NoteTagSource::Account(operator_account_id))
+            .map(|record| record.tag)
+            .collect();
```
TomasArrachea: We can simplify this to `NoteTag::with_account_target(operator_account_id)` and avoid the query to the store.

crates/faucet/src/lib.rs
ricomateo: Done here [a42961b](https://github.com/0xMiden/faucet/pull/292/commits/a42961b77eadb91e49f37250f69e2c7a7c897f16) (it required changing the test helper a little bit).

crates/faucet/src/lib.rs
ricomateo: Nice catch, fixed it here [f3b0fa6](https://github.com/0xMiden/faucet/pull/292/commits/f3b0fa6a07f141e1d87fa1dc43d944f98d4f31ab)

TomasArrachea (approved):

PR #306, opened 2026-09-21
feat(frontend): redesign the faucet page

dotsolo:
## Summary

Reskins the faucet frontend to the new Miden faucet design:

- new hero illustration, the Miden brand faces (Seasonmix for headings, labels and buttons, Söhne Buch for body), dark mode
- one recipient field with the wallet connect action as an inline pill
- a public / private segmented toggle with a single send button (previously two buttons); private is the default and the button stays inactive until the field holds a valid address
- the minting and result states render inline in the page column instead of as modal overlays
- issuance (with a ring meter) and the faucet address as two chips above the footer, and the official Miden wordmark in the footer

The behaviour is unchanged apart from the toggle. `ui.js` now wires the two tabs and the single send button (`setNoteType`) where it previously bound two separate buttons; `app.js`, `api.js` and `utils.js` are untouched. Every element id the controller looks up is preserved, so wallet connect, the PoW time hint, errors, minting, public / private completion and the note download + manual import steps all run through the existing code paths.

## Changes

- `index.html`: new markup. The result views are sections in the form column; the existing `.active` toggling shows them, and a `:has()` rule hides the form (and the minting view once a completed view is active) without any JS changes.
- `index.css`: rewritten around the design tokens, `prefers-color-scheme` dark mode, single-column layout at ≤560px. The two brand web fonts (taken from miden.xyz) are embedded as data URIs so the server needs no extra routes; that is most of the file's size.
- `ui.js`: public / private toggle, single send button with the inactive state, and the issuance meter driven through a CSS custom property instead of a bar width.
- `header.webp`: new hero illustration (1714×341, lossless WebP, 38 KB, down from 70 KB). Same filename, so no server changes.

## Notes for reviewers

- The design shows a "Paste" pill next to the address field; this keeps the wallet connect button there instead (as a "Connect" pill), since the app relies on the wallet adapter flow.
- The subtitle says "Free testnet tokens" rather than naming a symbol, because the token symbol is configured per deployment.
- Result screens keep the existing "Miden Wallet" wording that `app.js` sets, and keep the manual note download + import steps for when the note transport is unavailable.
- `:has()` needs Chrome 105+, Safari 15.4+ or Firefox 121+.
- The Google Fonts links are gone; the page no longer loads anything from a third party.

## Testing

Rendered every UI state headlessly through the real `UIController` (initial load, metadata loaded, private mode, validation error, wallet connected, minting, public complete, private complete with the note sent, private complete with manual download, after download with the warning, and load failure) at desktop and mobile widths and in dark mode, with no console or page errors.


TomasArrachea:
Hey! I just tried it locally to check how it's looking. One feedback: I think it would be nice if the page fitted in the screen without having to scroll down. We could scale things down so they look smaller. This is how I'm seeing it:


https://github.com/user-attachments/assets/0298c0de-42e8-4f44-a966-e150a04267ef

igamigo:
It would be great if the redesign contemplated the fact that private note minting is going to be removed soon (context: https://github.com/0xMiden/faucet/issues/307#issuecomment-5767494594)

dotsolo:
@TomasArrachea Great point. Let me fix that quickly and I'll reshare!

dotsolo:
@igamigo - Should it then just be "Send a note" without a public/private option and a note that says by default the note is private?

bobbinth:
> Should it then just be "Send a note" without a public/private option and a note that says by default the note is private?

Yes, this would work. I would maybe even just say "Send tokens" or something similar.

dotsolo:
Updated!

igamigo (approved):
Thanks! We'll fix the test and merge this

ricomateo (approved):

ricomateo:
Thanks! some commits were not signed so I force-pushed signing them so we can merge

Issue #307, opened 2026-09-21
Integrate funding service in the backend


igamigo:
Currently, the faucet backend works by maintaining the state of a network faucet operator account. This account either gets created and deployed, or gets imported from the network whenever the backend service starts. 

Considering this model will be incompatible with how the native token will be managed in future testnets and  mainnet (where the token will be bridged from Ethereum), we need to modify how it works. 

A new infra service is being introduced (https://github.com/0xMiden/node/issues/2561) which we might be able to reuse. This funding service's goal is to distribute assets from a single account to requestors. Additionally, the service implements some of functionality that the faucet also does. Namely, the funding service:

- Maintains an account that holds assets
- Syncs P2ID notes targeting the service account and periodically consumes them
- Receives requests of `(account_id, amount)` and returns a P2ID note as bytes with `amount` tokens
 aimed at `account_id` 

With this, the faucet could become mostly a gatekeeper in front of the funding service. It would no longer need to own an account, mint tokens, or build and submit transactions itself.

It would keep the PoW rate limiter and API keys, the `max_claimable_amount` and request complexity checks, the frontend, and the `/get_note` and `/send_note` endpoints for returning the note bytes from the service, either directly or through the NTL. For getting tokens, once the PoW challenge is validated, the faucet would call the funding service with `(account_id, amount)` and return the resulting note.

That would let us remove quite a bit:

- We can remove `miden-client` from the backend crate dependencies, unless we make use of the NTL client on the backend? I don't recall right now whether it's the backend or frontend's responsibility to send it over the NTL. On the frontend service we might still need the webclient for polling the notes, wallet adapter interfacing with the wallet, etc.
  - With this, the client SQLite store goes. API keys are stored there today, so those would need to move somewhere else. Not sure if some are settings are kept there
- The whole minting path: MINT notes, state sync, note screening, local/remote proving, and fee sponsorship.
- CLI token metadata flags such as `--token-symbol`, `--decimals`, and `--max-supply`, along with local issuance tracking. 

There are a few things we should clarify on the funding service side first:

- The faucet currently lets users choose public or private, so either the service needs to accept a note type or we drop the public option as suggested by @SantiagoPittella (would allow dropping even more code, but also eliminates that option for users altogether which I don't think is ideal). Supporting private notes on the funding service should be simple so I would opt for that.
- We need to decide how errors map back to the faucet: insufficient balance, service unavailable, amount above the service cap, etc. 
[...]

PR #308, opened 2026-09-21
refactor: integrate funding service

ricomateo:
This PR is the second step of the refactor described in #307.

Refactorizes the faucet with the minimal changes required to integrate with the funding service.

Introduces the following changes
- Upgrades the `miden-node-proto-build` dependency to `0.17.0-rc.1`.
    - This upgrade broke code that the refactor removes anyway, so this PR deletes it instead of updating it.
- Updates the `start` command to require a funding service URL along with token decimals.
- Updates the `/get_tokens` endpoint to forward the request to the funding service.

This PR leaves a lot of code pending to remove, but that will be addressed in follow-up PRs.

### How to test it

Start a node from the `rust-sdk` repository's root (make sure the version is at least 0.17.0-rc.1).

```bash
make start-node
```

Start the funding service from the `node` repository's root (make sure the version is at least 0.17.0-rc.1).

Make sure to set the `RUST_SDK_DIRECTORY` variable with the absolute path to the rust-sdk directory. The funder account is read from there.

```bash
# Set the path to the rust-sdk repository
export RUST_SDK_DIRECTORY=</path/to/rust-sdk>

cargo run --release -p miden-funding-service -- start \
  --listen 127.0.0.1:50401 \
  --rpc.url http://127.0.0.1:57291 \
  --tx-prover.url http://127.0.0.1:50051 \
  --account-file "$RUST_SDK_DIRECTORY/data/funders/wallet_15.mac" \
  --validator-signing-public-key 020c06515b355a62133ae98e53e4b5d3e6ee9ff60ce620a436780e4e308a3ff3e9
```

Run the following command to start the faucet (from the `faucet` directory in the branch `ricomateo-funding-service-integration`)

```bash
cargo run --release -p miden-faucet -- start \
  --network localhost \
  --funding-service-url http://127.0.0.1:50401 \
  --decimals 6
```

Navigate to http://localhost:8080 and send a token request


bin/faucet/src/main.rs
```diff
@@ -921,90 +806,6 @@ mod tests {
...
     }
 
-    // INTEGRATION TEST
-    // ---------------------------------------------------------------------------------------------
-
-    /// This test starts a stub node, a faucet connected to the stub node, and a chromedriver
-    /// to test the faucet website. It then loads the website, mints tokens, and checks that all the
-    /// requests returned status 200.
-    #[tokio::test]
-    async fn frontend_mint_tokens() {
```
ricomateo: this has been removed for now but will be added later in a follow up PR

bin/faucet/src/funding_service_client.rs
```diff
@@ -0,0 +1,252 @@
...
+use crate::COMPONENT;
+
+// CLIENT
+// ================================================================================================
+
+/// How long the faucet waits for a funding request.
+///
+/// The funding service answers only once the note is committed, which takes at least one block
+/// interval, so the node timeout does not apply here. The service's own HTTP timeout is 5 minutes,
+/// past which it answers 408; the faucet waits up to 1 minute.
```
TomasArrachea: Actually, the funding service returns the note optimistically, it does not wait for it to be committed to respond. So we could use the same timeout for all requests.

bin/faucet/src/main.rs
```diff
@@ -65,13 +57,11 @@ const ENV_POW_BASELINE: &str = "MIDEN_FAUCET_POW_BASELINE";
 const ENV_BASE_AMOUNT: &str = "MIDEN_FAUCET_BASE_AMOUNT";
 const ENV_ENABLE_OTEL: &str = "MIDEN_FAUCET_ENABLE_OTEL";
 const ENV_STORE: &str = "MIDEN_FAUCET_STORE";
+const ENV_DECIMALS: &str = "MIDEN_FAUCET_DECIMALS";
 const ENV_EXPLORER_URL: &str = "MIDEN_FAUCET_EXPLORER_URL";
-const ENV_BATCH_SIZE: &str = "MIDEN_FAUCET_BATCH_SIZE";
+const ENV_FUNDING_SERVICE_URL: &str = "MIDEN_FAUCET_FUNDING_SERVICE_URL";
```
TomasArrachea: Let's apply these changes also to the example .env and docs, either on this PR or the followup.

bin/faucet/src/api/get_metadata.rs
```diff
@@ -25,25 +24,36 @@ pub struct Metadata {
 
 #[instrument(parent = None, target = COMPONENT, name = "server.get_metadata", skip_all)]
 pub async fn get_metadata(State(server): State<ApiServer>) -> Json<GetMetadataResponse> {
+    // The balance is read per request so the page shows what is left right now. A funding service
+    // that cannot be reached leaves it out rather than failing the whole page.
+    let balance = match server.funding_service.status().await {
```
TomasArrachea: The metadata endpoint has no auth and runs on every frontend load, so we should probably add a short-lived cache here so we don't forward all requests to the funding service.

bin/faucet/src/main.rs
```diff
@@ -147,6 +116,15 @@ pub enum Command {
...
         config: ClientConfig,
 
+        /// Base URL of the funding service that emits the notes.
+        #[arg(long = "funding-service-url", value_name = "URL", env = ENV_FUNDING_SERVICE_URL)]
+        funding_service_url: Url,
+
+        /// Decimals of the token the funding service hands out, used by the frontend to convert
+        /// base units into token amounts.
+        #[arg(long = "decimals", value_name = "U8", env = ENV_DECIMALS)]
+        decimals: u8,
```
TomasArrachea: Would be nice if the funding service provided the token decimals (maybe along the token name) in the status endpoint, so we don't need to set them manually here cc @SantiagoPittella

SantiagoPittella: hmmm,  it might be a good idea. Also, we could get this from the node, right?

TomasArrachea: Yes, but the idea is to remove the miden-client dependency here, so we won't have the RPC client to connect to the node.

SantiagoPittella: We could just create a client for the node, but I get why we might want to avoid that, and just relay on the funding service. Is there anything else asset-related that we could want to read from the funding service?

ricomateo: it would be useful if the funding service includes the token symbol of the native asset in the `/status` endpoint response

igamigo: I commented on this in the original issue: https://github.com/0xMiden/faucet/issues/307#issuecomment-5763868917. We need to expose issuance, address, metadata about the account in general.

SantiagoPittella: I opened https://github.com/0xMiden/node/pull/2688

SantiagoPittella: I included decimals, symbol and name. Which are static and are read at startup. Issuance can change overtime, for this we will need to include some periodic refetching

igamigo: > Issuance can change overtime, for this we will need to include some periodic refetching

I think that's the idea. Or, why would it be a problem? IIUC Mirko was saying this would be nice to show so that people don't need more direct access to the funding service: https://github.com/0xMiden/faucet/issues/307#issuecomment-5763913127
EDIT: maybe remaining funds is enough for this? I was imagining enabling displaying the same progress bar that we already display, but perhaps remaining funds is enough

bin/faucet/src/main.rs
ricomateo: Updated the .env and docs here [a9a16a1](https://github.com/0xMiden/faucet/pull/308/commits/a9a16a10151c3cfa8bcb0fd09b834ce6443b03b5)

bin/faucet/src/api/get_metadata.rs
ricomateo: Done here [f5a9b2a](https://github.com/0xMiden/faucet/pull/308/commits/f5a9b2ac5088a67f044dcd4e2d510fd729f2c8a4)

bin/faucet/src/funding_service_client.rs
ricomateo: Nice catch, fixed it and updated the doc comments here [0ab017e](https://github.com/0xMiden/faucet/pull/308/commits/0ab017e79972901377382a10411a0e2592c29bb0)

igamigo (approved):

PR #309, opened 2026-09-21
chore: remove support for private notes

ricomateo:
**Motivation**

This PR is the first step of the refactor described in #307, after which the faucet will no longer support private notes.

The refactor requires updating dependencies to 0.17.0-rc.1, which touches a lot of code. Removing private notes first, in this separate PR, avoids updating code that would be deleted anyway and keeps the refactor PR smaller.

This is how the frontend looks now

image


igamigo (approved):
LGTM! But would be nice if @TomasArrachea could take a look

TomasArrachea (approved):
LGTM

PR #310, opened 2026-09-23
chore: remove dead code

ricomateo:


igamigo (approved):
LGTM! Let's create an issue to see if we want to keep the `miden-faucet-client` crate

